PR #35, opened 2015-07-25
Fix daylight cycle

BeYkeRYkt:
Fixed:
* gamerule "doDaylightCycle" for 1.5.2

Fix based on: http://wiki.vg/Protocol#Time_Update


Shevchik:
This will actually make daynight cycle work regardless of rules. Is this intended?

BeYkeRYkt:
Maybe, but it's the only way I can see to fix this problem. And like in 1.5.2 are no negative values of time.

Shevchik:
Okay, i will check it later, there should be a way to stop daynight cycle/

PR #65, opened 2015-10-14
Fix crash with invalid enchantments in 1.5.2/1.6.X

MrPowerGamerBR:
I just copied and pasted the 1.8 enchant verification to remove the enchantments from 1.5.2 and 1.6.X clients, didn't test, but maybe it will work (Maybe with a bit of tweaking).

Will test later.


Shevchik:
It won't even compile version is an enum, not a string

MrPowerGamerBR:
Yeah, I forgot to put a .toString() before the .startsWith()

Or I could simply used getVersion() == ProtocolVersion.MINECRAFT_1_7_5 || getVersion() || ProtocolVersion.MINECRAFT_1_7_10

Shevchik:
Probably there should be a separate registry for that, like with block/item remapping

PR #72, opened 2015-10-31
[MCPE] Remap Stained Glass (95) to Glass (20)

MrPowerGamerBR:
If not remapped, a invisible block is created on a MCPE client.

Also @Shevchik, good luck on the MCPE branch! It is progressing more and more with every commit, I remember way back, when I couldn't even connect with the MCPE branch, and now, I can connect and even interact with the world! Soon you will be able to *play* on a Bukkit Server with a MCPE client :smile:  :+1:


Shevchik:
That remapper is really outdated, it only contains 0.9 items and should be updated.

MrPowerGamerBR:
Yeah, I also forgot to add a remap on the Item Remapper, because, as of now, Stained Glass are shown as a ".name" stone.

PR #73, opened 2015-11-01
[MCPE] Move block and item (unused?) to unified remapper, add new remapping.

MrPowerGamerBR:
~~TODO: Move all block remapping to the unifed remapper (maybe not needed? most of the blocks in the old remapper are outdated)~~ The old unifed remapping was outdated, I'm now creating a new remap list.

~~TODO: Change MINECRAFT_PE to MINECRAFT_PE_0_12~~ (view Shevchik comments below)

TODO: Move the rest of the old remappers

Also, I removed the item remapper class without any problems, is it unused?

This change also fixes the "update block" on tall grasses :smile: However, I "think" this needs more testing before pushing to the mcpe branch, but I could connect and the block remapping (it is only for the stained glass -> glass) is working.


Shevchik:
Yep, i do not remap items for mcpe yet. Also renaming pe version is not needed, i won't support more than one anyway.

MrPowerGamerBR:
So you will only support the latest version of MCPE? Makes sense (well, nobody runs a outdated MCPE client just to play on Servers)

Shevchik:
Yep, mcpe autoupdates automatically anyway.

MrPowerGamerBR:
Also, maybe a SubID remapper will be needed? Because, after analyzing the MCPE IDs (and the MC PC IDs), some blocks have a different SubID and ID.

Example: Acacia logs are in the ID 17 on MCPE with a different SubID (I think it is 4), however, in Minecraft PC, they have ID 162 and SubID 0.

MrPowerGamerBR:
If only the latest version will be supported, then how the "BEFORE_MINECRAFT_PE_0_13" should be named?

MrPowerGamerBR:
If all the default remappers are removed, almost every block starts showing correctly, without "Update Block!" (acacia logs, acacia leaves, tall grass (and every other flower), etc etc etc).

I only kept some remappers (stained glass -> glass, stained glass pane -> glass pane, redstone wire -> carpet, redstone torch on/off -> torch) and I will add another remappers after I find inexistent items in MCPE.

MrPowerGamerBR:
Also @Shevchik, you should change the default skin to Steve instead of Alex, because you are sending a Alex skin, however MCPE client is rendering the Steve model, so if you see near the player arms, there will be a line of black pixels (due to the Alex skin having the arm smaller than the Steve model)

Shevchik:
I m just sending skin which was sent by my  client when i started. The

Shevchik:
Problem is probably because of slim variable being wrong. Should switch it to true.

src/protocolsupport/protocol/transformer/mcpe/packet/mcpe/clientbound/ChunkPacket.java
```diff
@@ -7,13 +7,18 @@
...
 import protocolsupport.protocol.transformer.mcpe.packet.mcpe.PEPacketIDs;
-import protocolsupport.protocol.transformer.mcpe.utils.BlockIDRemapper;
+import protocolsupport.protocol.typeremapper.id.IdRemapper;
+import protocolsupport.protocol.typeremapper.id.RemappingTable;
 import protocolsupport.utils.MutableBlockPosition;
 
 public class ChunkPacket implements ClientboundPEPacket {
-
+	private static final RemappingTable blockRemapper = IdRemapper.BLOCK.getTable(ProtocolVersion.MINECRAFT_PE);
+	private static final RemappingTable itemRemapper = IdRemapper.ITEM.getTable(ProtocolVersion.MINECRAFT_PE);
```
Shevchik: Remove that, we don't need item, entity, mapcolor remapper in chunk packet

src/protocolsupport/utils/ProtocolVersionsHelper.java
```diff
@@ -27,6 +27,10 @@
 		ProtocolVersion.MINECRAFT_1_5_2
 	};
 
+	public static final ProtocolVersion[] BEFORE_MINECRAFT_PE_0_13 = new ProtocolVersion[] {
```
Shevchik: Just use ProtocolVersion.MCPE

src/protocolsupport/protocol/transformer/mcpe/handler/ClientboundPacketHandler.java
```diff
@@ -92,7 +93,11 @@
 import net.minecraft.server.v1_8_R3.World;
 
 public class ClientboundPacketHandler {
-
+	private static final RemappingTable blockRemapper = IdRemapper.BLOCK.getTable(ProtocolVersion.MINECRAFT_PE);
```
Shevchik: All of them should be protected, or they will be copied to every anonymous class in transformers registry

MrPowerGamerBR:
@Shevchik made all suggestions you made :+1:

MrPowerGamerBR:
Yay! Merged!

Later I will start looking on what blocks/items are missing to remap, I "think" I already remapped every missing block, I just need to remap the missing items.

PR #74, opened 2015-11-03
[MCPE] Update (and add new) block remappings for MCPE.

MrPowerGamerBR:
New block remappings for MCPE.

Almost every block (not item, I will do the item remapping when item remapping is working for the MCPE client, so I can test it) remapping is done for MCPE.

Also, some blocks can't be remapped (example: Podzol: MC 1.8.8 ID: 3:1; MCPE ID: 243) and some maybe missing or wrong.

The problem is that the client desyncs from the server a lot, walking? desync. viewing a unremapped block? desync. doing nothing? desync. (And, when it does desync, the only fix is disconnecting the client and kicking it from the console), maybe those desyncs will be fixed as the mcpe branch goes more stable and feature complete :smile: 

Also, IDs are different, I was getting (and comparing) every ID on Minecraft Wiki, however, then I found that some IDs are wrong (when comparing in the PocketMine-MP project: https://github.com/PocketMine/PocketMine-MP/blob/mcpe-0.12/src/pocketmine/item/Item.php#L73 ), so maybe some remappings are wrong.

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/id/IdRemapper.java
+++ b/src/protocolsupport/protocol/typeremapper/id/IdRemapper.java
@@ -112,6 +112,8 @@ public class IdRemapper {
 			// (sticky) piston -> stone
 			registerRemapEntry(29, 1, ProtocolVersionsHelper.MCPE);
 			registerRemapEntry(33, 1, ProtocolVersionsHelper.MCPE);
+			// piston head -> stone
+			registerRemapEntry(34, 1, ProtocolVersionsHelper.MCPE);
 		}
 		@Override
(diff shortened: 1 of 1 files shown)


Shevchik:
You can easily desync if your connection is poor. ProtocolSupport currently doesn't implement RakNet protocol fully, so it just handles whatever packets are received without reordering and waiting for missed ones.

MrPowerGamerBR:
@Shevchik I'm testing on my local network, so I think it is not because of the connection.

Also, when I say "desync", it is because the MCPE clients stops reciveing/sending packets, example: connected players stops moving, chunks doesn't load, etc.

PR #124, opened 2016-01-13
Fix rabbit hide id

nathampa0909:
Fix 415 (rabbit hide) to 334 (leather).


Shevchik:
Rabbit hide is not direct equivalent to leather. Especially in crafting. So i am not sure if we really want to do this.

nathampa0909:
But 4 rabbit hide = 1 leather
http://www.minecraftinfo.com/leather.htm

MrPowerGamerBR:
So 4 "leathers" will become 1 leather.

That's a bit confusing if you are crafting.
Em 13 de jan de 2016 12:18, "Nathan Guedes" <notifications@github.com>
escreveu:

PR #251, opened 2016-08-27
Polar bear now looks like a sheep

tylerhasman:
I believe that a polar bear is more close to a sheep than a spider. You may argue that players might be confused why a sheep is attacking them. That is a good point.


Shevchik:
Yea, that's why it is a spider. Attacking sheep will really look strange.

PR #261, opened 2016-09-12
chmod +x ./gradlew

Maxqia:


Shevchik:
Use tabs for indentation.

Maxqia:
Well eclipse is giving me an error for it, https://docs.oracle.com/javase/7/docs/api/java/util/Iterator.html

games647:
@Maxqia Do you have the JDK 7 or JDK 8 installed? I assume it's, because you still have JDK 7 installed. Java 8 has declared Iterator.remove() as default method.

Maxqia:
I have JDK 8, but it's set to 1.7 compatibility mode by the current build.gradle
https://github.com/ProtocolSupport/ProtocolSupport/blob/716e52a923705f42e9788ac92a9b33e2c396299d/build.gradle#L9

games647:
@Maxqia Often this tries to find JDK 7 and if it doesn't find it, it changes only the bytecode version. New libraries could still be used although they don't exist in Java 7

Shevchik:
The remove method is default in java8, and i'm using java8 to build it. I'm going to drop java7 support soon, so it doesn't ammter anyway.

Maxqia:
Can you bump the java version now then?

PR #282, opened 2016-10-09
Fixes for some stuff

Maxqia:
This fixes the RecyclableSingletonList from #261 and fixes MylesIsCool/ViaVersion#449

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ plugins {
 group 'protocolsupport'
 version '4.25.dev'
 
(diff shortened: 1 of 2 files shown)


Shevchik:
Encoder is not supposed to write to output.

Maxqia:
The 1.10 decoder used to write to the output before the "Encode 1.10 packets using middle packets" commit, anyways edited my commit to remove the ChannelHandlerContext stuff, but it didn't seem to update this PR because it's closed now.

PR #285, opened 2016-10-11
Support the legacy custom effect splash potion.

EntryPointKR:


Shevchik:
According to what i see, CustomPotionEffect tag is supported starting from 1.7.
So i think a transformer for that should be separate.

Shevchik:
Will merge for now, will investingate CustomPotionEffects tag and its versions later.

PR #287, opened 2016-10-11
Fixes for some stuff (Again)

Maxqia:
This fixes the RecyclableSingletonList from #261 and fixes MylesIsCool/ViaVersion#449 (Again). I edited the commit to only write to the output buffer.


Shevchik:
Still nope, encoder stays as it is now, because middle packets are used.

PR #288, opened 2016-10-11
Update #36 for latest ProtocolVersion

BeYkeRYkt:
Check #36 for information about this request


Shevchik:
This packet doesn't have a player, because it is not marked as needsPlayer one.
Also trusting client disconnect message is a bad idea anyway, some plugins read logs and make assumtions based on disconnect reason.

PR #291, opened 2016-10-14
Update 1.5.2 health display

EntryPointKR:


Shevchik:
needsPlayer() should be used only when really needed because the access is async. Are you sure that max health is not sent anywhere so we can cache it?

EntryPointKR:
What's means "cache it"?
This is my best in my understanding for ProtocolSupport.
If the wrong this code, please point out.

Shevchik:
ProtocolSupport has a thing called LocalStorage. It caches data that may be needed by another packet later. See entity spawn and entity metadata transformers.

EntryPointKR:
Like this?

Shevchik:
No, i will commit the proper fix later.

Issue #297, opened 2016-10-29
Please don't remove /reload for paper users


RoboMWM:
In paper, I use the `/reload permissions` command, but since protocolsupport doesn't know this... it removes it, so yea..

PR #298, opened 2016-10-30
Use PlayerCommandPreprocessEvent to block reloads

Zbob750:
Switches away from removing the command from the server command map, to using Bukkit's API to block usage. This method is less likely to break should someone make changes to the command map.

Fixes GH-297 in the process.


Shevchik:
Wtf is that block code, you don't need to iterate to over list here, use a hashset.
Also this won't work if more spaces are added between /reload and arguments, bukkit will parse it fine, but preprocess event receives original message.

Zbob750:
Was trying to keep it simple. Will switch it to a hashset and look into spaces issue.

Zbob750:
Updated, spaces are a non-issue. Feel free to check.

Zbob750:
Easiest way to maintain case insensitivity is with a TreeSet, to do it with a HashSet is going to result in a lot of overengineering just to save an extremely tiny amount of time.

Shevchik:
Or you can just toLowerCase() on message before checking if it exists in the map.

Zbob750:
I really need to pull my head out of my ass tonight. Thanks for pointing that out >.> I didn't even think of it.

Shevchik:
Now just fix the formatting, this project uses tabs, not spaces.

Zbob750:
Woops, fixed

PR #387, opened 2017-02-16
Fix Books created in 1.8 have broken content

EntryPointKR:
1.8 seems **\n** is next line, **\\\\n** and **\\"** is string.


Shevchik:
This should be in a serializer

Shevchik:
A separate method for transforming read itemstack should be used (see writeItemStack). Also that str format looks some sort of unescape. The escape method of 1.8.8 client should be investigated and full unescape method should be written i think.

EntryPointKR:
I can't check the version because [here](https://github.com/ProtocolSupport/ProtocolSupport/blob/9230f5b7c9d8f552f1944e0b2cb746030bcb1d92/src/protocolsupport/protocol/packet/middleimpl/serverbound/play/v_1_8__1_9_r1__1_9_r2__1_10__1_11/CustomPayload.java#L14) ProtocolVersion.getLatest()

Shevchik:
Okay, i will show you what i meant tomorrow.

Shevchik:
Okay, looks like 1.8 encodes client->server book pages as json text component, so a proper conversion would be parsing text as text component and getting the text value.

Issue #363, opened 2017-01-24
Books created in 1.8 have broken content


Brokkonaut:
If you create a book in 1.8 and sign it, newlines appear as "\n" in the text and some characters display as "\u003d" and similar and there are double quotes around the whole text.

MrPowerGamerBR:
Can confirm, 1.8.9:
![http://i.imgur.com/jq84Pek.png](http://i.imgur.com/jq84Pek.png)
![http://i.imgur.com/QMB1Q0e.png](http://i.imgur.com/QMB1Q0e.png)
![http://i.imgur.com/GH33CjF.png](http://i.imgur.com/GH33CjF.png)

Shevchik:
Books in 1.8.9 are written using json components, while at later versions they are just raw strings.

Shevchik:
Should be fixed now

EntryPointKR:
@Shevchik Crash when book open

[06:00:22 INFO]: EntryPoint lost connection: Internal Exception: io.netty.handler.codec.DecoderException: protocolsupport.libs.com.google.gson.JsonSyntaxException: protocolsupport.libs.com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 3 path $
[06:00:22 INFO]: EntryPoint left the game

Shevchik:
Which version?

EntryPointKR:
Client 1.8.9
Server 1.11.2 PaperSpigot

Shevchik:
Should be fixed now

EntryPointKR:
1.8 is fixed but 1.4 ~ 1.7 is crash.

Shevchik:
Should be fixed now.

PR #388, opened 2017-02-17
Fix broken book content issue #363

EntryPointKR:


EntryPointKR:
"full unescape method" mean is this?

PR #389, opened 2017-02-17
Fix broken book content issue

EntryPointKR:
@Shevchik 
Does "full unescape method" mean this?


Shevchik:
Yeah, but the method is still in the wrong place. I'm refactoring this part of the serializer ATM, so this pull will have to wait.

PR #402, opened 2017-03-03
use the official port for mcpe

caspervanneck:


MrPowerGamerBR:
I'm 100% sure Shevchik knows what is the official MCPE port and I'm 100% sure he set it to 2222 for a reason.

Shevchik:
I set it to whatever port that is not default for PC and PE. There is a reason behind that, mainly because currently i expect people to actually check the source or wait before using it.

NycuRO:
@Shevchik Can you add this costumizable?

DaMatrix:
ProtocolSupport doesm't have a config file, and it isn't going to be changed until this is production-ready. If you want to use it on production, have a look at linux `stone`.

PR #428, opened 2017-03-20
Head Rotation + Weather Packets + Gamemode Update

7kasper:
Fixed the yaw clipping/jumping of pc clients' heads by applying a
constant.

### DIFF
--- a/src/protocolsupport/protocol/legacyremapper/pe/PELevelEvent.java
+++ b/src/protocolsupport/protocol/legacyremapper/pe/PELevelEvent.java
@@ -1,0 +1,103 @@
+package protocolsupport.protocol.legacyremapper.pe;
+
+import protocolsupport.api.ProtocolVersion;
+import protocolsupport.protocol.packet.middleimpl.ClientBoundPacketData;
+import protocolsupport.protocol.serializer.MiscSerializer;
+import protocolsupport.protocol.serializer.VarNumberSerializer;
+
+/***
+ * 
+ * Information gathered from the Nukkit project :)
+ *
+ */
+public class PELevelEvent {
+    public static final int EVENT_SOUND_CLICK = 1000;
+    public static final int EVENT_SOUND_CLICK_FAIL = 1001;
+    public static final int EVENT_SOUND_SHOOT = 1002;
+    public static final int EVENT_SOUND_DOOR = 1003;
+    public static final int EVENT_SOUND_FIZZ = 1004;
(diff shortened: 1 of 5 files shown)


Shevchik:
So the pe yaw 256 degrees based?

7kasper:
It seems like in the pc packet it's 256 degrees while pocket expects 360 degree based. When I look at the yaw in F3 on the pc it is also based on 360 degrees, but while debugging the packet it never reached 180 degrees.

NycuRO:
You can fix FOV?

NycuRO:
<3

7kasper:
@NycuRO Is there anything wrong with the FOV? It only bugs while running, but I believe @DaMatrix fixed that.

NycuRO:
@7kasper FOV isn't able to change..

NycuRO:
@7kasper And chests isn't usable... Enchantments... and other. Can you fix that? Or metadata?

7kasper:
@NycuRO I enjoy your enthusiasm :) 

I have already implemented three things today and have to learn for an upcomming tests, but I'm sure we'll implement those features in some time. You'll have to wait though.

NycuRO:
I need for SkyWars that :> @7kasper

gdude2002:
@NycuRO Those requests are kinda out of scope of this pull request. Please give the developer some time.

@7kasper Apologies, they've been doing this on the Discord as well.

MrPowerGamerBR:
@7kasper Shevchik said he won't implement inventories until Mojang adds server sided inventories to MCPE, because the current MCPE inventory system sucks.

7kasper:
@gdude2002 you're right. There's alot left to do. You know what the funny thing is though, protocol support seems more stable already than most pocket serversoftwares I tried :P

gdude2002:
To expand on the inventory stuff, it's pretty much entirely client-side. Unfortunately, that means some of the packets you would need for a conventional implementation just don't exist - most of the time, you don't even know what GUI is open.

Mojang has stated that they are working on a client-side inventory, though.

Shevchik:
Weather and gamemode change looks fine, still unsure about head rotation, this needs more investigation. Can you split those pull requests?

7kasper:
@Shevchik There are no errors, but if you open the current implementation, you can see the heads of pc players jump around 128 degrees (to -128 degrees) on pocket edition clients.

Shevchik:
I was sure that mojang only used 256 degree computation for entity teleport, not player. I guess they screwed it up.

PR #432, opened 2017-03-23
Send PC skins to PE players

DaMatrix:
Feel free to close this if you want, just thought somebody might find this useful.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ libs/
 *.iml
 /dllibs/
 classes/
+run/
--- a/src/protocolsupport/protocol/legacyremapper/pe/PESkin.java
+++ b/src/protocolsupport/protocol/legacyremapper/pe/PESkin.java
@@ -4,6 +4,8 @@ import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.concurrent.Callable;
 
 import javax.imageio.ImageIO;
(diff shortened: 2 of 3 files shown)


jojoe77777:
And if the PC player's skin is in slim (Alex) format?

DaMatrix:
good point, didn't think of that :P
it would just show some extra rows of black pixels on the arms

Shevchik:
You can't do it like that, it will block an netty io thread.

7kasper:
It's funny. I was working on the same thing. Here is my version:
`	public static final byte[] getSkin (String username){
		URL imageLocation;
		try {
			imageLocation = new URL("http://s3.amazonaws.com/MinecraftSkins/" + username + ".png");
		} catch (MalformedURLException e1) {
			return STEVE;
		}
		return new Callable<byte[]>() {
			@Override
			public byte[] call() {
				try {
					return toNetworkData(ImageIO.read(imageLocation));
				} catch (Exception e) {
					return STEVE;
				}
			}
		}.call();
	}`

Of course you'll also need to change the validator to accept old format skins:
`Validate.isTrue(skin.getHeight() == 64 || skin.getHeight() == 32 , "Must be 32 or 64 pixels high");`
Pocket still handles the 64*32 images fine.

DaMatrix:
Same thing though, it still downloads the image on the netty thread, locking it.

7kasper:
@DaMatrix I wasn't saying it was anything different :)

DaMatrix:
The other thing i can think of would be reading from the SkinsRestorer cache files, assuming that the plugin is present. (not a good fix though :/ )

7kasper:
So we'll need to stall the particular packet untill the skin is downloaded on another thread and then callback on it someway.

PR #448, opened 2017-04-02
[PSPE] Entity interaction

7kasper:
Implemented entity hitting and interaction Client -> Server

I can't test interaction that yet because, I think, pocket will only allow for interaction when it knows it is interactable (IE a boat or villager) which is send with the EntityMetadata *ugh*.

So modern (PC) entity interaction (right-click) works with two packets. One older packet which is just what entity (ID) is clicked and a newer packet that also tells where the entity is clicked (primarely used for armour stands). Unless the entity is an armour stand, both packets are sent. Otherwise only the latter newer packet (interactAt) is sent. Pocket doesn't have an interactAt yet, so we don't know where an entity is clicked precisely, but I think that will be implemented when armorstands come to pocket edition. For now I just send an empty vector so plugins using the new packet (and bukkit event) will (probably) work.


src/protocolsupport/protocol/packet/middleimpl/serverbound/play/v_pe/Interact.java
```diff
@@ -0,0 +1,53 @@
...
+	public RecyclableCollection<ServerBoundPacketData> toNative() {
+		RecyclableArrayList<ServerBoundPacketData> packets = RecyclableArrayList.create();
+		WatchedEntity target = cache.getWatchedEntity(targetId);
+		switch(peAction){
+			case INTERACT: {
+				if(target.getType() != SpecificRemapper.ARMOR_STAND){
+					packets.add(MiddleUseEntity.create(targetId, MiddleUseEntity.Action.INTERACT, null, 0));
+				}
+				packets.add(MiddleUseEntity.create(targetId, MiddleUseEntity.Action.INTERACT_AT, new Vector(), 0)); //TODO: Send where the entity is clicked (will probably be implemented with armorstands.)
+			}
```
duchuyctlk: If there is no `break` at the end of a `case` block, the statement for the next cases will be executed also.

7kasper: @duchuyctlk Thanks. I was fooled by the braces.

Shevchik:
Needs updating to latest version. Also don't add packet ids that are not implemented yet.

7kasper:
@Shevchik I updated it. Works again now 😄

PR #450, opened 2017-04-04
[PSPE] EntityStatus

7kasper:
Many things like hurt, dying or a wolf wobbling are sent in minecraft via entity events.
I've implemented all entity events currently available in MCPE. 

~~But the code **definitely** needs rework.
It needs to be in a remapper that maybe shouldn't even be pocket-specific (_these status codes might also change in the future_!)~~

~~Until we've got the entity meta remapper working and I can check for entitytypes less awkwardly, I don't want to make a fancy remapper for this, but this PR will be updated when due.~~

One little thing left:
It seems the fishing rod shooting is also handled in entitystatus, while the computer does this differently. (I think it just moves the dobber entity) So when implementing fishing we'll have to check if PE doesn't mind about that.


Shevchik:
Needs updating to latest version and converted to use a remapper and new entity types.

7kasper:
@Shevchik Updated. Works :)

It currently does not check for unknown status codes in PE, but I checked a few (PE) unimplemented ones and they don't seem to cause trouble. Makes the code more simple.

PR #453, opened 2017-04-09
Schism of SpecificRemapper and WatchedType

7kasper:
Already makes things less awkward.

For now the SpecificRemapper just transforms from WatchedType crudely and the metadata remapper still uses the SpecificRemapper, but that remapper has to be changed anyway.

I just copied most of the values of the SpecificRemapper, and added a few bukkit types.

### DIFF
--- a/src/protocolsupport/ProtocolSupport.java
+++ b/src/protocolsupport/ProtocolSupport.java
@@ -19,6 +19,7 @@ import protocolsupport.protocol.packet.handler.AbstractLoginListener;
 import protocolsupport.protocol.pipeline.initial.InitialPacketDecoder;
 import protocolsupport.protocol.typeremapper.id.IdRemapper;
 import protocolsupport.protocol.typeremapper.watchedentity.remapper.SpecificRemapper;
+import protocolsupport.protocol.typeremapper.watchedentity.types.WatchedType;
 import protocolsupport.protocol.typeskipper.id.IdSkipper;
 import protocolsupport.protocol.utils.data.ItemData;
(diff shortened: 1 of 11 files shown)


Shevchik:
Some formatting notes:
Add spaces between if and (
Add spaces between } (
Use tabs

7kasper:
@Shevchik Better? Some of the comments don't seem to show nicely in github but they look beautiful in eclipse :stuck_out_tongue:

Shevchik:
Just remove those comments, they are useless anyway (class outline shows everything anyway)

7kasper:
@Shevchik Developers always seem to have fueds over formatting. I like the comments especially in big classes. Whatever.

gdude2002:
@7kasper Those were pretty useless comments, I agree with @Shevchik - comments should explain _why_ you're doing what you're doing,  not _what_ you're doing. If you're wondering what the code is doing, well.. read the code.

7kasper:
@gdude2002 It's personal preference. If I come back to old projects I find it very useful if I can quickly see what a class does and where it does that. But I don't mind adapting of course :relaxed:

Shevchik:
Also you can't use the EntityType to get object network ids, they are different.

7kasper:
@Shevchik What? Are the bukkit ids different from the protocol? Why?
If that is the case I'll put the differing ids back, but I will get something to eat first.

Shevchik:
Because the bukkit ids are the nbt save ids. And network object ids are different from that.

PR #456, opened 2017-04-11
Quick playerMetadata fix.

7kasper:
You should probably delete or flag build 40.
I messed up a bit yesterday. The player doesn't even have an id so the temporary function in specificremapper didn't give any errors but skipped the player (as it has an id of -1).

Derp


Shevchik:
I have already rewritten this part, will push it after testing

PR #518, opened 2017-06-16
Implement "SpawnPainting" packet

MrPowerGamerBR:


7kasper:
Cool :) So is this also called when you load a chunc with a painting?

MrPowerGamerBR:
@7kasper yup

PR #519, opened 2017-06-16
Implement "SpawnExpOrb" packet

MrPowerGamerBR:
To avoid code redundancy I created a static method in SpawnGlobal since the code used in XP orbs and lightning are the same. (except the entity type)

However this may or may not be a problem when entity metadata is implemented. (since, according to @DaMatrix, the XP count for XP orbs are sent in the metadata (instead of PC where it is sent in the SpawnExpOrb packet))

### DIFF
--- a/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/SpawnExpOrb.java
+++ b/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/SpawnExpOrb.java
@@ -1,0 +1,17 @@
+package protocolsupport.protocol.packet.middleimpl.clientbound.play.v_pe;
+
+import protocolsupport.api.ProtocolVersion;
+import protocolsupport.protocol.packet.middle.clientbound.play.MiddleSpawnExpOrb;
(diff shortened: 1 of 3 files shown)


7kasper:
It's a neat hack, but it seems weird to call the todata from another packet. I think both this and the spawnglobal which I did previously should be calling a common method in another class. Like the way it is done in the PELevelEvent class. Maybe just PESpecialEntitys.createPacket? You can add the the entityIds as public static final ints.

Shevchik:
We don't need entity types separately, only the mapping pc -> pe, so placing create in addEntity is fine.

7kasper:
@Shevchik Since all these spawn packets are handled by just one PE packet, if that method changes you would have to change the same code in multiple scripts.
But yes. It will get harder to navigate, I think we need just one spawnclass that spawnliving also uses.

Shevchik:
Conflicts, needs to be updated

MrPowerGamerBR:
@Shevchik will fix as soon as possible 👍

MrPowerGamerBR:
@Shevchik fixed 😄

PR #521, opened 2017-06-16
[PSPE] UseItem / Block interaction

7kasper:
Implemented useItem (Right-Clicking)

As @MrPowerGamerBR said, PSPE is now the first minecraft pockets software to fully support redstone 😃


MrPowerGamerBR:
@7kasper and it supports some quirks that are not even available in vanilla MCPE! 😛

src/protocolsupport/protocol/packet/middle/serverbound/play/MiddleBlockPlace.java
```diff
@@ -20,6 +20,7 @@
 
 	@Override
 	public RecyclableCollection<ServerBoundPacketData> toNative() {
+
```
Shevchik: Remove that pls

PR #534, opened 2017-06-25
[PSPE] Entity Move & Look

7kasper:
Changed the way entityMove works and added entityLook & headRotation.
It was a bit difficult to figure everything out, but in the end it was just caching, transforming and rehashing.
All PC move packets are now send to PE aswell instead of relying on entityTeleport. The network data cache is needed to store the position, rotation and headrotation of entities since PC packets only send what needs to be updated and PE expects everything. The system seems to work farely well and entities feel alot less static now 😄 

I'm not quite sure, but maybe the entity injector thingy isn't needed anymore. (At least for special PE packets)


Shevchik:
This still doesn't account for entity velocity, and there is no way to do that. This might lead to position sync issues.

7kasper:
@Shevchik I thought so too, but I think once in a certain amout of ticks the server sends a teleport packet is aswell. You can sometimes see an entity jump a small part of a block, but when you're not actively looking for it you don't notice.

Shevchik:
Not every server does this.

7kasper:
So disabling the entitytracker does allow PSPE to be run on Paperspigot. You can see the entities correcting from time to time. I'ts a bit more than with regular spigot but still barely noticeable. I've run some test and it seems to work well until there are a lot of entities (50+). Also whenever lag occurs a huge backlog can cause entities to sputter around until reaching their final destinations. 
If anyone else wants to test, go right ahead. I'll try to keep this PR up to date.

MrPowerGamerBR:
@7kasper I think when Shevchik said

> Not every server does this.

He meant server implementations (like Glowstone) doesn't send the teleport packet after a few ticks.

7kasper:
@MrPowerGamerBR Well... I think you might want to.  Sync problems can probably occur on normal clients too. Too much vector changes or some time lag will cause that. There's a reason why servers send these absolutes in packets. 

I actually tested glowstone, but when punching pig it sends it flying across the universe :P (Probably no resetting of the velocity)

But I don't think this PR will make a difference in it anyway. Either we inject into spigot to give more entityteleport packets (the current solution) or listen to all minecraft packets and cache the information to adjust changes. Pitch, Yaw and headYaw are send in absolutes anyway and I think (might be wrong here though) only the deviding in the entity velocity packet for pocket gives rise to the sync problems that need to be dealt with.

7kasper:
_I am guilty!_
Did some more testing together with entityMetadata today. I tried and completely removed the injector.
At first it seemed fine like I said earlier, but after a bit of harder testing entities started teleporting around all over the place and alot too. It does depend on your server implementation, but this can't work yet.

I'm guessing Mojang will not change their movement system for Minecraft java.
Either we leave the injecor, but edit it to also do entityLook and headYaw. (I got this to work aswell).
`Just a shame that we need to rely on NMS and Spigot heavy code.`
Or we need to write some movement calculation thread. Perhaps an async task that requests the current position from the server. (I've just seen that this is the way BKCommonLib does it) I think that'd be a bit more laggy than the current way, but less implementation specific.

7kasper:
I implemented something very experimental.. A task executed each tick that calculates an entitie's position based on the velocity and sends a RelMove to PE each tick. Testers are welcome! On my faster PC it looks good, but than again it also looks good without the entityTracker and the motionUpdater. On my older PC this experiment looks worse.

Shevchik:
This task will literally kill the server performance.

Asuna-Yuuki:
I can't test it I have mcpe 1.2 😢😭

7kasper:
Perhaps someone can implement entityLook and headYaw packets from this and make a new branch to make them work with the old system (maybe you need to add them in the entitytracker too...) but those packets send absolutes..

NycuRO:
@7kasper Ready PR?

MrPowerGamerBR:
No, this isn't finished, he said that in ProtocolSupport's Discord channel (and it needs to be rewritten).

NycuRO:
And why them not rewritten?

MrPowerGamerBR:
Because nobody starting rewritting it yet, more important things need to be implemented than just entity move/look. (which is just a cosmetic thing, I doubt someone would complain that a entity isn't rotating it's head)

7kasper:
@MrPowerGamerBR I would xD I see this zombie attacking me and it is looking the wrong way! I wan't to feed sheep and it doesn't even look at my face. But you're right about the more important things :wink:

7kasper:
I'm going to close this. With the changes in #750 animals will move like they should and although it is platform specific it impacts performance much less.

PR #543, opened 2017-07-07
/protocolsupport command improvements

xNaXDy:
CommandHandler:
- Moved (args.length == 1) into one enclosing if statement
- Added proper usage message in case of invalid command. onCommand now
always returns true.
- Added checks for new config values (see below)

ProtocolSupport:
- Added two config values
- expose-plugin if set to false will send unknown command message if
someone without permission attempts to execute /ps to avoid exposure of
the plugin if wanted. default value is true (send "no power" message)
- list-only-relevant-versions if set to true will only list versions
that have connected players in /ps list ; default value is false (list
all versions regardless)

These changes should make ProtocolSupport commands more accessible to
users without programming knowledge.

### DIFF
--- a/src/protocolsupport/commands/CommandHandler.java
+++ b/src/protocolsupport/commands/CommandHandler.java
@@ -29,7 +29,7 @@ public class CommandHandler implements CommandExecutor, TabCompleter {
 	@Override
 	public boolean onCommand(CommandSender sender, Command cmd, String label, String[] args) {
 		if (!sender.hasPermission("protocolsupport.admin")) {
-			sender.sendMessage("Unknown command. Type \"/help\" for help.");
+			sender.sendMessage(ChatColor.RED + "You have no power here!");
 			return true;
 		}
 		else if(args.length == 1 || args.length == 2) {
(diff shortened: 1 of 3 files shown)


Shevchik:
No config values are allowed

xNaXDy:
Fixed it and added an optional parameter value to /ps list

Shevchik:
Usage could actually be moved to plugin.yml command description

src/protocolsupport/commands/CommandHandler.java
```diff
@@ -38,12 +38,12 @@ else if(args.length == 1 || args.length == 2) {
 				return true;
 			}
 			else if (args[0].equalsIgnoreCase("list")) {
-				for (ProtocolVersion version : ProtocolVersion.values()) {
-					if (version.isSupported()) {
+				for (ProtocolVersion version : ProtocolVersion.getAllSupported()) {
+					//if (version.isSupported()) {
```
Shevchik: remove the code, not just comment it

xNaXDy:
I haven't done proper testing with this build yet, which is why I only commented it for now. I'll commit another final version later today for you to review

xNaXDy:
I've added two more things and removed the comments. Should be final now.

PR #554, opened 2017-07-17
[Patch] Fire damage

7kasper:
Implemented top-level reset on the way Bukkit deals with fire damage
using an event listener.

My guess is that since minecraft 1.12 fireDamage is send not via the convential methods anymore (ie. EntityStatus) and is handled by the client.  (I tested versions before and it doesn't send entityStatus)
I noticed that Entities in versions before 1.11 actually didn't damage from anything firerelated at all.
I thought it would be simple enough to just send the damage hardly through bukkit if anything fire related happens, so a low level eventhandler could take care of that.


7kasper:
You might want to wait a bit. This PR will solve the issue, but minecraft 1.12 players will not receive the firedamage sound anymore, even with the damage cause.

Shevchik:
This should send needed packets only for affected versions.

7kasper:
@Shevchik Indeed. I was a bit too optimistic that the damgeCause would just make the correct sound play. That'd be an easy fix. O well we just have to figure out what distance players can hear sounds. Bleh.

7kasper:
I ended up picking 48 blocks. Seems like a pretty good number. Currently using the distanceSquared method provided by Bukkit's locations. It seems however that normal minecraft calculates these things in boxes (Like if you're within ~48, ~256, ~48) If you want that I can make a method for it, but the effect is not so noticable from that far and you can't even hear the sounds anymore.

Although the player now sees the damage being dealt from fire, sound effects for other players are still missing when they are on fire, but not in fire. Hmm

src/protocolsupport/listeners/PlayerListener.java
```diff
@@ -70,6 +72,25 @@ public void onBlockPlace(BlockPlaceEvent event) {
 			connection.sendPacket(ServerPlatform.get().getPacketFactory().createBlockBreakSoundPacket(new Position(block.getX(), block.getY(), block.getZ()), block.getType()));
 		}
 	}
+	
+	@EventHandler(priority = EventPriority.LOWEST, ignoreCancelled = true)
+	public void onEntityDamage(EntityDamageEvent event) {
+		if((event.getCause() == DamageCause.FIRE_TICK || event.getCause() == DamageCause.FIRE)) {
+			for(Player p : Bukkit.getOnlinePlayers()) {
```
Shevchik: Iterate over player world players, not all.

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_9r1_9r2/WorldSound.java
```diff
@@ -14,8 +14,10 @@
 
 	@Override
 	public RecyclableCollection<ClientBoundPacketData> toData(ProtocolVersion version) {
+		String soundname = LegacySound.getSoundName(id);
 		ClientBoundPacketData serializer = ClientBoundPacketData.create(ClientBoundPacket.PLAY_WORLD_CUSTOM_SOUND_ID, version);
-		StringSerializer.writeString(serializer, version, LegacySound.getSoundName(id));
+		if(soundname == "entity.player.hurt_on_fire") soundname = "entity.player.hurt";
```
Shevchik: What the? A unified remapper should be used if multiple versions need sound names remapping.

src/protocolsupport/listeners/PlayerListener.java
```diff
@@ -76,10 +78,9 @@ public void onBlockPlace(BlockPlaceEvent event) {
 	@EventHandler(priority = EventPriority.LOWEST, ignoreCancelled = true)
 	public void onEntityDamage(EntityDamageEvent event) {
 		if((event.getCause() == DamageCause.FIRE_TICK || event.getCause() == DamageCause.FIRE)) {
-			for(Player p : Bukkit.getOnlinePlayers()) {
-				//Send extra status packets to all old connections within 48 blocks (2304 = 48^2).
-				if(p.getLocation().distanceSquared(event.getEntity().getLocation()) <= 2304) { 
-					Connection connection = ProtocolSupportAPI.getConnection(p);
+			for(Entity e : event.getEntity().getWorld().getNearbyEntities(event.getEntity().getLocation(), 48, 128, 48)) {
```
Shevchik: That's quite ineffective actually, the server itself has a method that allows getting nearby entities by class in a more effective way. Maybe that method should be added to PlatformUtils.

7kasper: @Shevchik Doesn't CraftBukkit call that server method using getNearbyEntities?
[Source](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/src/main/java/org/bukkit/craftbukkit/CraftWorld.java#706)

src/protocolsupport/zplatform/impl/spigot/SpigotMiscUtils.java
```diff
@@ -181,5 +194,28 @@ public void setFraming(ChannelPipeline pipeline, IPacketSplitter splitter, IPack
...
+	@Override
+	public List<Connection> getNearbyConnections(Location loc, double deltaX, double deltaY, double deltaZ, ProtocolVersion... versions) {
+		return getNearbyCraftEntities(loc, deltaX, deltaY, deltaZ, SpigotMiscUtils.playerFilter()).map(e -> ProtocolSupportAPI.getConnection((Player) e.getBukkitEntity())).filter(PlatformUtils.legacyConnectionFilter(versions)).collect(Collectors.toList());
+	}
+	
+	@Override
+	public List<Player> getNearbyPlayers(Location loc, double deltaX, double deltaY, double deltaZ) {
+		return getNearbyCraftEntities(loc, deltaX, deltaY, deltaZ, SpigotMiscUtils.playerFilter()).map(e -> (Player) e.getBukkitEntity()).collect(Collectors.toList());
+	}
+
```
7kasper: Like this? 😄

NycuRO:
@7kasper Thanks for PR. 👍

PR #559, opened 2017-07-19
Remap 159 (Stained Clay) to 35 (Wool)

junilima:
Both blocks are colored, used as decoration, the buildings will look better than stones.


MrPowerGamerBR:
But both blocks doesn't have the same breaking time, so that will cause issues when someone tries to break a block.

junilima:
I did this on my own and had no problems, so I decided to suggest the move to the next versions, so I did not need to change constantly.

Shevchik:
No, the reason is already said above.

PR #561, opened 2017-07-19
[PSPE] SpawnObject

7kasper:
Adds the spawnobject for pocket.

**Items:**
Items not fully implemented yet.
So PC sends a spawnObject packet for objects which includes Items.
This spawnObject packet however does not include itemdata whatsoever. The ID, Damage and such are send with the first EntityMetadata packet.
PE on the other hand expects all itemdata when you summon the item via the AddItem packet. There is no way to update an item later on with a SetEntityData packet.

So the implementation goes half way. First on spawn we prepare an itemEntity without the item information. And with the entityMetadata (Which I'm still working on) we set the ItemStack and send the AddItem packet for PE and remove it from the cache.

I need the SpawnObject packet to go trough with EntityMetadata. But I thought to PR this seperately first.

From [leaky](https://github.com/Hydreon/Steadfast2/tree/master/src/pocketmine/network/protocol/v120) I don't see any changes in 1.2 as how to Item Entities are spawned. They might never change it, as even though behaviour is different to PC, it is Server based and superiour than always sending two packets to spawn an item (Like PC does)


Shevchik:
Yeah, but the main problem actually is not this one. The main problem is that you can't dynamically update the itemstack. I guess it will be that one part where we will have to rely on server sometimes syncing the real position.

7kasper:
I'll test entityHanging tomorrow. But since items can't be dynamically updated, we need to wait for all the information ( & entitymetadata) and then spawn it in. I'll add that with my metdata PR, but that isn't finished yet. (It's kinda big :P)

MrPowerGamerBR:
@7kasper don't waste your times with item frames, item frames will be changed to entities (like PC) in 1.2 :wink:

PR #565, opened 2017-07-22
Update concrete remapping

jordinjm:
Map concrete to stained clay.

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/id/IdRemapper.java
+++ b/src/protocolsupport/protocol/typeremapper/id/IdRemapper.java
@@ -20,7 +20,7 @@ public class IdRemapper {
 
 	public static final IdRemappingRegistry<ArrayBasedIdRemappingTable> BLOCK = new IdRemappingRegistry<ArrayBasedIdRemappingTable>() {
 		{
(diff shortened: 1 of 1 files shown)


half-cambodian-hacker-man:
+1 - Excessive use of concrete on a map when ProtocolSupport is used can make the world look like utter ~~shit~~ garbage for players on versions older than 1.12.

AppleDash:
+1

Shevchik:
Concrete hand break time - 9units, stained clay - 6.25. Which means most of the time player won't be actually break a block.

MrPowerGamerBR:
Also, if you don't care about hand break time (and your players raging because they can't break concrete) then use ProtocolSupport's API to remap concrete to stained clay.

jordinjm:
Oh I didn't realize they had different breaking times. I will just make a plugin for it.

7kasper:
@jordinjm Your best bet is to make a plugin that uses the remapcontroller and gives slowness to legacy players that are trying to break concrete. That way you can seemingly affect the hardness. It just has to be a higher breaktime than concrete and the server will do the rest. If you match the time up correctly, it won't be a nuisance for the player.

PR #570, opened 2017-07-23
[PSPE] Explode with particles and sound.

7kasper:
I started with TNT. The other things just followed. (An explosion you can't hear or see is no explosion 😄)

**Explode**
PE needs three packets for it. The explode, but also a expel motion for the player the packet is send to and particle effect, because the PE client doen't do that on its own.

**Particles**
Send by PELevelEvent. Still I needed extra id-remapping- and skipping-table + some things involving count and offset.

**WorldSound**
Expands on @MrPowerGamerBR 's custom sound. I saw that the first sound name in the Sound_Definitions.json of MCPE is the same as Minecraft 1.9 so I just added ProtocolVersion.MINECRAFT_PE to all 1.9 remaps.


src/protocolsupport/protocol/typeremapper/pe/PELevelEvent.java
```diff
@@ -56,10 +53,9 @@
 	public static final int SOUND_SPLASH = 3506;
 	public static final int CAULDRON_TAKE_WATER = 3507;
 	public static final int CAULDRON_ADD_DYE = 3508;
-
+
```
Shevchik: ws

src/protocolsupport/protocol/typeremapper/pe/PELevelEvent.java
```diff
@@ -75,10 +71,51 @@
 	public static final int SOUND_CAULDRON_FILL_WATER = 3506;
 
 	public static final int SET_DATA = 4000;
-
+
```
Shevchik: ws

PR #571, opened 2017-07-23
[Patch] Readded unimplemented sounds, becasue some WHERE implmented :F

7kasper:


7kasper:
Something went wrong here? o_O

Shevchik:
I already did this.

PR #579, opened 2017-07-26
[PSPE] Break animations

7kasper:
It's much hacking again but it works okay.
Not perfect. But okay.

Timing will never be perfect. Since we are kind of limited to ticks there will always be a change we're just hitting one tick off or maybe there's some network lagg. This will make it so the animation finishes too early or late. I think this is however the best we can do.

### DIFF
--- a/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/BlockBreakAnimation.java
+++ b/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/BlockBreakAnimation.java
@@ -1,0 +1,21 @@
+package protocolsupport.protocol.packet.middleimpl.clientbound.play.v_pe;
+
+import protocolsupport.api.ProtocolVersion;
+import protocolsupport.protocol.packet.middle.clientbound.play.MiddleBlockBreakAnimation;
+import protocolsupport.protocol.packet.middleimpl.ClientBoundPacketData;
+import protocolsupport.protocol.typeremapper.pe.PELevelEvent;
+import protocolsupport.utils.recyclable.RecyclableCollection;
+import protocolsupport.utils.recyclable.RecyclableSingletonList;
+
+public class BlockBreakAnimation extends MiddleBlockBreakAnimation {
+
+	@Override
+	public RecyclableCollection<ClientBoundPacketData> toData(ProtocolVersion version) {
+		if (stage >= 0 && stage <= 9) {
+			return RecyclableSingletonList.create(PELevelEvent.createPacket(PELevelEvent.BLOCK_START_BREAK, position, stage));
+		} else {
+			return RecyclableSingletonList.create(PELevelEvent.createPacket(PELevelEvent.BLOCK_STOP_BREAK, position));
+		}
+	}
+
+}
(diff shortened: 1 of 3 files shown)


src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/BlockBreakAnimation.java
```diff
@@ -0,0 +1,68 @@
...
+
+	private final TIntObjectHashMap<TimePle> animation = new TIntObjectHashMap<>();
+	
+	@Override
+	public RecyclableCollection<ClientBoundPacketData> toData(ProtocolVersion version) {
+		if (stage >= 1 && stage <= 9) {
+			if (!animation.containsKey(entityId)) {
+				animation.put(entityId, new TimePle(false, System.nanoTime(), stage));
+			} else {
+				if (!animation.get(entityId).getStarted() && stage > animation.get(entityId).getStartStage()) {
```
Shevchik: extract get to variable

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/BlockBreakAnimation.java
```diff
@@ -0,0 +1,68 @@
...
+					animation.get(entityId).setStarted(true);
+					return RecyclableSingletonList.create(PELevelEvent.createPacket(PELevelEvent.BLOCK_START_BREAK, position, (int) (
+							65535 / Math.ceil(((((System.nanoTime() - animation.get(entityId).getStartTime()) / (stage - animation.get(entityId).getStartStage())) * (9 - animation.get(entityId).getStartStage())) / 1000000000) * 20 + 12.5)
+						)));	
+				}
+			}
+		} else {
+			if (animation.containsKey(entityId)) {
```
Shevchik: unneeded check

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/BlockBreakAnimation.java
```diff
@@ -0,0 +1,68 @@
...
+import protocolsupport.utils.recyclable.RecyclableSingletonList;
+
+public class BlockBreakAnimation extends MiddleBlockBreakAnimation {
+
+	private final TIntObjectHashMap<TimePle> animation = new TIntObjectHashMap<>();
+	
+	@Override
+	public RecyclableCollection<ClientBoundPacketData> toData(ProtocolVersion version) {
+		if (stage >= 1 && stage <= 9) {
+			if (!animation.containsKey(entityId)) {
```
Shevchik: Use extracted get variable and check for null instead of contains

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/BlockBreakAnimation.java
7kasper: What do you mean?

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/BlockBreakAnimation.java
```diff
@@ -11,25 +11,24 @@
...
-			if (!animation.containsKey(entityId)) {
-				animation.put(entityId, new TimePle(false, System.nanoTime(), stage));
+			TimePle animation = animations.get(entityId);
+			if (animation == null) {
+				animations.put(entityId, new TimePle(false, System.nanoTime(), stage));
 			} else {
-				if (!animation.get(entityId).getStarted() && stage > animation.get(entityId).getStartStage()) {
-					animation.get(entityId).setStarted(true);
+				if (!animation.getStarted() && stage > animation.getStartStage()) {
+					animations.get(entityId).setStarted(true);
```
Shevchik: Also reuse extracted variable here

7kasper: Indeed. I'll also adjust the calculation a bit. Seems that slow mining blocks can be calculated much more arurately and maybe shouldn't suffer under the extra constant.

PR #598, opened 2017-08-03
Add command support using slashes, remove old command prefixes

MrPowerGamerBR:
This implements the "Request Command" packet, which, since 1.2, allows commands not sent via the available command packet to be executed.

While this code works:tm:, the code within `MiddleChat#create(...)` is so protocol dependent that it almost hurts inside when reading it, I tried finding any other examples of other MiddleXYZ packets that has that method implement but almost none of them has it. (and the ones that has it doesn't need the version object) so I wonder if that's the proper way to do it. 👍

Anyway, this reads the request command packet and creates a MiddleChat packet for the server to process.

This commit also removes the old command prefixes `mcpenew` branch had to execute commands, since this commit obsoletes that old "hack".

### DIFF
--- a/src/protocolsupport/protocol/packet/middle/serverbound/play/MiddleChat.java
+++ b/src/protocolsupport/protocol/packet/middle/serverbound/play/MiddleChat.java
@@ -1,5 +1,7 @@
 package protocolsupport.protocol.packet.middle.serverbound.play;
 
+import protocolsupport.api.ProtocolType;
+import protocolsupport.api.ProtocolVersion;
 import protocolsupport.protocol.packet.ServerBoundPacket;
 import protocolsupport.protocol.packet.middle.ServerBoundMiddlePacket;
 import protocolsupport.protocol.packet.middleimpl.ServerBoundPacketData;
@@ -19,4 +21,9 @@ public abstract class MiddleChat extends ServerBoundMiddlePacket {
 		return RecyclableSingletonList.create(creator);
 	}
 
+	public static ServerBoundPacketData create(String message) {
+		ServerBoundPacketData creator = ServerBoundPacketData.create(ServerBoundPacket.PLAY_CHAT);
+		StringSerializer.writeString(creator, ProtocolVersion.getLatest(ProtocolType.PC), message);
+		return creator;
+	}
 }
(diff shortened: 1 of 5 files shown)


src/protocolsupport/protocol/packet/middleimpl/serverbound/play/v_pe/CommandRequest.java
```diff
@@ -0,0 +1,31 @@
...
+import protocolsupport.protocol.packet.middleimpl.ServerBoundPacketData;
+import protocolsupport.protocol.serializer.StringSerializer;
+import protocolsupport.protocol.serializer.VarNumberSerializer;
+import protocolsupport.utils.recyclable.RecyclableCollection;
+import protocolsupport.utils.recyclable.RecyclableSingletonList;
+
+public class CommandRequest extends ServerBoundMiddlePacket {
+
+	protected String command;
+	protected int type;
```
Shevchik: No need to store data that we don't use

src/protocolsupport/protocol/packet/middle/serverbound/play/MiddleChat.java
```diff
@@ -19,4 +21,9 @@
 		return RecyclableSingletonList.create(creator);
 	}
 
+	public static ServerBoundPacketData create(String message) {
+		ServerBoundPacketData creator = ServerBoundPacketData.create(ServerBoundPacket.PLAY_CHAT);
+		StringSerializer.writeString(creator, ProtocolVersion.getLatest(ProtocolType.PC), message);
```
Shevchik: ProtocolVersionsHelper.LATEST_PC

src/protocolsupport/protocol/packet/middle/serverbound/play/MiddleChat.java
```diff
@@ -19,4 +21,9 @@
 		return RecyclableSingletonList.create(creator);
 	}
 
+	public static ServerBoundPacketData create(String message) {
```
Shevchik: The toNative method should use this one

MrPowerGamerBR:
@Shevchik done, also, should the packet read every value from it (even if ProtocolSupport doesn't use it) or should the code just skip the remaining bytes?

Shevchik:
Read every value

MrPowerGamerBR:
@Shevchik then it is done 👍 (I only forced pushed a new commit to remove the old ProtocolVersion import from the MiddleChat class)

PR #611, opened 2017-08-08
Avoid gc stressing MessageFormat.format call

winofi:
Avoid MessageFormat.format call in protocolsupport.api.ProtocolVersion.OrderId.compareTo because it produces huge amounts of characters (600kB/player/second on our spigot servers). Maybe a simple "Cant compare order from different types" or "Cant compare order from different types: this - " + type + ", other - " + o.type will be sufficient to avoid the format call.

image

Image shows 5 Minutes of jmc Flight Recording on ~10 Player Spigot server.


Shevchik:
That will make exception message unclear sometimes, but performance is more important in this case.
Should also check other places, and switch to using ifs or lambdas later.

PR #633, opened 2017-08-12
[PSPE] start the raknet server on the same ip as the mc server

jfmherokiller:
this change makes the raknet server start on the same ip/interface as the one defined in the config file of the server.

### DIFF
--- a/src/protocolsupport/zplatform/impl/glowstone/injector/GlowStoneNettyInjector.java
+++ b/src/protocolsupport/zplatform/impl/glowstone/injector/GlowStoneNettyInjector.java
@@ -74,7 +74,11 @@ public class GlowStoneNettyInjector {
 	public static void startPEServer() {
 		try {
 			GameServer gameserver = getGameServer();
-			peserver = new RakNetServer(new InetSocketAddress(PENetServerConstants.TEST_PORT), PENetServerConstants.PING_HANDLER, new UserChannelInitializer() {
+			String serverIp = gameserver.getServer().getIp();
+			if (serverIp.isEmpty()) {
+				serverIp = "0.0.0.0";
+			}
+			peserver = new RakNetServer(new InetSocketAddress(serverIp,PENetServerConstants.TEST_PORT), PENetServerConstants.PING_HANDLER, new UserChannelInitializer() {
 				@Override
 				public void init(Channel channel) {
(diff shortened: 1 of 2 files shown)


Asuna-Yuuki:
this is a bad idea

MrPowerGamerBR:
Probably it is right, (I didn't test this PR yet) the current RakNetServer is binding to 0.0.0.0 while the Java server is binding to the IP setted in server.properties when the RakNetServer should respect the IP in server.properties.

MrPowerGamerBR:
(whoops, misclicked the "Comment" button)

Shevchik:
Now just need to check if it works if the configured ip is empty

MrPowerGamerBR:
Or I'm doing something wrong or something is wrong in that commit, I can't connect to the MCPE server, even I set the server-ip to "0.0.0.0", to nothing and even if I set it to my local IP address.

Maybe I'm doing something wrong...

jfmherokiller:
@MrPowerGamerBR you are not doing something wrong i forgot to fix the code so that it will default to the wildcard address if it cant resolve the address defined by the config file

src/protocolsupport/zplatform/impl/glowstone/injector/GlowStoneNettyInjector.java
```diff
@@ -74,7 +74,12 @@ public static void inject() throws IllegalArgumentException, IllegalAccessExcept
 	public static void startPEServer() {
 		try {
 			GameServer gameserver = getGameServer();
-			peserver = new RakNetServer(new InetSocketAddress(PENetServerConstants.TEST_PORT), PENetServerConstants.PING_HANDLER, new UserChannelInitializer() {
+			String serverIp = gameserver.getServer().getIp();
+			if(serverIp.isEmpty())
```
Shevchik: space between if and (

src/protocolsupport/zplatform/impl/glowstone/injector/GlowStoneNettyInjector.java
```diff
@@ -74,7 +74,12 @@ public static void inject() throws IllegalArgumentException, IllegalAccessExcept
...
+			{
```
Shevchik: Should be on the same line with if (don't forget the space)

PR #639, opened 2017-08-14
Fix Unnecessary qualified reference

jfmherokiller:


Shevchik:
No point, next change in ps master branch will change the version again. Which will reuire manual merge again.

PR #642, opened 2017-08-16
remap concrete to wool

lukechatton:
Wool better matches the new concrete block for builds.


MrPowerGamerBR:
But wool doesn't have the same block breaking time as concrete, so no.

Kinda duplicate of https://github.com/ProtocolSupport/ProtocolSupport/pull/565

If you don't care about players saying "oh I can't break this wool (which actually is a concrete block) this server is laggy!!!" then use ProtocolSupportStuff to remap everything that you want to: https://www.spigotmc.org/resources/protocolsupportstuff.44527/

PR #651, opened 2017-08-19
Add sub ID remapping for items, fix some remaps due to ID differences…

MrPowerGamerBR:
… and some small fixes

I didn't like how "ItemIdToPEIdSpecificRemapper" is kinda messy, but I don't know a better way to improve that part of the code.

And I also created a `registerBlockAndItemRemap` method to keep PEDataValues tidy. (so we don't need to register the same block two times, one for the block remap and another for the item remap)

I also fixed the fence remapper, because it seems Mojang swapped the dark oak/acacia data values.

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/itemstack/toclient/ItemIdToPEIdSpecificRemapper.java
+++ b/src/protocolsupport/protocol/typeremapper/itemstack/toclient/ItemIdToPEIdSpecificRemapper.java
@@ -12,21 +12,14 @@ public class ItemIdToPEIdSpecificRemapper implements ItemStackSpecificRemapper {
 	public ItemStackWrapper remap(ProtocolVersion version, String locale, ItemStackWrapper itemstack) {
 		IntTuple itemAndData = PEDataValues.ITEM_ID.getRemap(itemstack.getTypeId(), itemstack.getData());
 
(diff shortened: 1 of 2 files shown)


src/protocolsupport/protocol/typeremapper/itemstack/toclient/ItemIdToPEIdSpecificRemapper.java
```diff
@@ -3,13 +3,32 @@
...
 
 public class ItemIdToPEIdSpecificRemapper implements ItemStackSpecificRemapper {
 
 	@Override
 	public ItemStackWrapper remap(ProtocolVersion version, String locale, ItemStackWrapper itemstack) {
-		itemstack.setTypeId(PEDataValues.ITEM_ID.getRemap(itemstack.getTypeId()));
+		IntTuple itemAndData = PEDataValues.ITEM_ID.getRemap(itemstack.getTypeId(), itemstack.getData());
+
+		if (itemAndData == null) { // If itemAndData is null, then we are going to get the "-1" remapper
+			// When a remapper uses -1, it means it is a "remap every sub ID to this new ID"
```
Shevchik: No need for comments like those here. Also with 1.13 this map will be gone, so no point to commenting the Complex remapper itself too

PR #663, opened 2017-08-27
[PE] Send crafting recipes v2

DaMatrix:
I deleted the old repo when i refactored my github profile, resubmitting this


7kasper:
Just a question. When I try crafting a torch.. Do I get 4 of them?

PR #669, opened 2017-09-03
Remap Vex and Evoker for MCPE clients

MrPowerGamerBR:


7kasper:
Next time, you can just commit that kind of stuff if you like 😛

PR #670, opened 2017-09-03
Remap spawn eggs for MCPE clients

MrPowerGamerBR:

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/itemstack/toclient/MonsterEggToPEIdSpecificRemapper.java
+++ b/src/protocolsupport/protocol/typeremapper/itemstack/toclient/MonsterEggToPEIdSpecificRemapper.java
@@ -1,6 +1,6 @@
 package protocolsupport.protocol.typeremapper.itemstack.toclient;
 
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.bukkit.entity.EntityType;
 import protocolsupport.api.ProtocolVersion;
 import protocolsupport.protocol.typeremapper.itemstack.ItemStackSpecificRemapper;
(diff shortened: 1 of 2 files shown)


src/protocolsupport/protocol/typeremapper/itemstack/toclient/MonsterEggToPEIdSpecificRemapper.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class MonsterEggToPEIdSpecificRemapper implements ItemStackSpecificRemapper {
+
+	@Override
+	public ItemStackWrapper remap(ProtocolVersion version, String locale, ItemStackWrapper itemstack) {
+		NBTTagCompoundWrapper tag = itemstack.getTag();
+		if (tag.isNull()) {
+			return itemstack;
+		}
+		String id = MinecraftData.removeNamespacePrefix(tag.getCompound("EntityTag").getString("id"));
```
Shevchik: This may fail if for whatever reason egg didn't have EntityTag

MrPowerGamerBR: @Shevchik wouldn't the same thing happen for pre-1.8 clients? After all, the code for MonsterEggToPEIdSpecificRemapper is exactly the same as the MonsterEggToLegacyIdSpecificRemapper, except the code to transform to PE ID of course.

But if you want it, I could add the tag check.

Shevchik: In your case the removeNamespacePrefix will fail, pc monster egg doesn't have it

PR #695, opened 2017-09-18
[PSPE] Dimtest

7kasper:
Need a eventlistener for some inventory stuff aswell I'm afraid.

This is just something very simple I just threw together. It seems to work, though debate about the necessity and approach is required. 
This is of course a simple thing that can also be included in PSPS if wanted.


Shevchik:
We don't do hacks like that. If dimension change doesn't work, we just omit it and that's all.

PR #710, opened 2017-10-04
Fix clients below 1.9 hearing pickup sounds globally

RainoBoy97:
Since commit https://github.com/ProtocolSupport/ProtocolSupport/commit/c9c462a393b60981fcbd6d69418d8c284cef6b5b players using a client version below 1.9 have heard item pickup sounds for all players due to the sound being played at their location, this commit fixes that issue by playing the sound at the item location.

This fix has been tested on a server running the latest version of Spigot (1.12.2-R0.1-SNAPSHOT) and ProtocolSupport (4.28-dev) with another player (1.12.2) and myself (1.7.10).


Shevchik:
The problem is that you can't actually query entity just like that, that's an async access that potentially won't work. Instead i think the collect effect should only play if pickup entity target id matches the player one.

RainoBoy97:
Alright. I guess I should leave that up to anyone with more experience with ProtocolSupport as I just looked at it to fix this issue 😛

PR #723, opened 2017-10-21
Fix UnsupportedOperationException under Paper

MrPowerGamerBR:
🗞

Something I did just to see if it was possible, so *maybe* the code sucks... a lot.

The reason ProtocolSupportPE's throws UnsupportedOperationException under Paper is because `trackedPlayers` is actually `trackedPlayerMap`'s keySet(), so, instead of editing `trackedPlayers` we need to edit the `trackedPlayerMap`.

We also need to set the `tracker` in the NMS Entity.

To avoid code duplication, `PaperEntityTrackerEntry` extends `SpigotEntityTrackerEntry`, that's why some of the fields/methods are now public.

While I didn't do extensive tests, it seems there isn't any issues when spawning entities/players.

One of the reasons I don't want this to be merged yet is because now ProtocolSupportPE would be compiled against Paper, and I guess we shouldn't do that. (but I think there isn't any other alternatives)


7kasper:
🤔

MrPowerGamerBR:
@7kasper 🤷‍♀️

Well, this fix does work but, you know, it would require compiling against Paper instead of Spigot, that's why I didn't merge it yet

src/protocolsupport/zplatform/impl/spigot/entitytracker/SpigotEntityTrackerEntry.java
```diff
@@ -105,8 +105,8 @@
 	private boolean isPassenger;
 	private boolean onGround;
 
-	private final Set<EntityPlayer> trackedDefaultPlayers = new HashSet<>();
-	private final Set<EntityPlayer> trackedPEPlayers = new HashSet<>();
+	public final Set<EntityPlayer> trackedDefaultPlayers = new HashSet<>();
```
ysl3000: Please make them protected. It's not the solution to allow access from outside of the class.

src/protocolsupport/zplatform/impl/spigot/entitytracker/SpigotEntityTrackerEntry.java
```diff
@@ -105,8 +105,8 @@
...
+	public final Set<EntityPlayer> trackedPEPlayers = new HashSet<>();
```
ysl3000: Same here

src/protocolsupport/zplatform/impl/spigot/entitytracker/SpigotEntityTrackerEntry.java
```diff
@@ -566,7 +566,7 @@ public void clear(final EntityPlayer entityplayer) {
 		}
 	}
 
-	private void addTrackedPlayer(EntityPlayer player) {
+	public void addTrackedPlayer(EntityPlayer player) {
```
ysl3000: here too.

src/protocolsupport/zplatform/impl/spigot/entitytracker/SpigotEntityTrackerEntry.java
```diff
@@ -576,7 +576,7 @@ private void addTrackedPlayer(EntityPlayer player) {
 		}
 	}
 
-	private void removeTrackedPlayer(EntityPlayer player) {
+	public void removeTrackedPlayer(EntityPlayer player) {
```
ysl3000: same here

build.gradle
```diff
@@ -38,7 +38,7 @@ gen.mkdirs();
 
 task updateLibs(type: Download) {
 	src([
-		'https://yivesmirror.com/files/spigot/spigot-1.12.2-R0.1-SNAPSHOT-b1408.jar',
+		'https://yivesmirror.com/files/paperspigot/PaperSpigot-1.12.2-b1244.jar',
```
ysl3000: You don't need it, when accessing fields with reflection.

Do you execute methods that needs it as a dependency?

build.gradle
```diff
@@ -102,7 +102,7 @@ repositories {
 }
 
 dependencies {
-	shadow files('buildprocessor/BuildProcessor.jar', new File(dllibs, 'spigot-1.12.2-R0.1-SNAPSHOT-b1408.jar'), new File(dllibs, 'glowstone-2017.10.0-20171007.064505-9.jar'))
+	shadow files('buildprocessor/BuildProcessor.jar', new File(dllibs, 'PaperSpigot-1.12.2-b1244.jar'), new File(dllibs, 'glowstone-2017.10.0-20171007.064505-9.jar'))
```
ysl3000: see first comment.

build.gradle
DaMatrix: reflection is slow, although that would work. there's got to be a better way though.

MrPowerGamerBR: @ysl3000 not everything uses Reflection, that's why it needs to be compiled against Paper

MrPowerGamerBR: ...to be honest I'm not even sure if it actually used any Paper stuff

ysl3000: That wasn't the intention of this my comment.

At one point you're using reflection. The rest of your code seems not to access any variables/methods that requires Paperspigot.

MrPowerGamerBR: Yeah, I need to double check if it can be compiled against Spigot, I think I was compiling against Paper for testing purposes (and because I was decompiling the NMS classes via IntelliJ IDEA's class decompiler) and I never switched back to Spigot to test if it would actually compile against it.

If it does compile, I will fix that and the other stuff you mentioned 😉

ysl3000: The intention was to show you that the code doesn't need paperSpigot to be executed.
So it can be merged, when changes are done.
Revert the dependency changes.

MrPowerGamerBR: @ysl3000 yeah I know, that's what I was saying.

I just never compiled against Spigot because I thought my code had some Paper specific stuff, but it seems that isn't the case after relooking the code.

MrPowerGamerBR: After fighting gradle because for some random reason it wasn't finding the RakNetServer dependency, it seems that it *does* use Paper specific stuff

https://i.imgur.com/N98Ho7S.png

However it can be replaced with Reflection

ysl3000: Looking into it

7kasper:
Take a look at refectionutils. Maybe you can try getting a field reference on object creation and use it to invoke methods on the map.

ysl3000:
@7kasper jep, will add it in the evening. Have do do some stuff before.

ysl3000:
[__reflected_not_tested.patch.txt](https://github.com/ProtocolSupport/ProtocolSupport/files/1486022/__reflected_not_tested.patch.txt)

MrPowerGamerBR:
I will try to fix this PR later, currently busy with other IRL stuff.

btw @ysl3000, if you want to create your own:tm: PR, go ahead! (you can use the code I already made for this PR if you want to, I don't care because, in the end, it will be pulled to PSPE anyway 😉)

ysl3000:
I’m busy as well.

MrPowerGamerBR:
I will fix this PR after the new changes from `master` are merged to `mcpenew` (a lot of changes :S)

CasperVanNeck:
when wil this be merged

MrPowerGamerBR:
When I have time to fix this branch to clean up and fix some misc issues

MrPowerGamerBR:
#755

PR #745, opened 2017-11-13
[PSPE] Fix for BB & rounded position collision :wink:

7kasper:
The problem: PE sends rounded positions (based on the change it does itself). This is really irritating as it intersects with boundingboxes or so the server thinks.

The way the fix works is it fakes a position apology to the server if the server sends a teleport that is within the leniency. This leniency is standard 0.5 blocks, which makes you just able to walk up a halfslab. When running up stairs of any sort (halfstairs, stairs cakeblocks etc etc), however, you sometimes skim through the BB of that for more than 0.5 blocks. So when you are moving up, the leniency of the position is set to 3 blocks.


Shevchik:
We might also just always fake apology to the server in all cases, because sometimes pe even has problems with normal blocks and such. And also it doesn't send the exact teleport locations back, so it is even hard to confirm the teleport sometimes.

7kasper:
Good point. So we fake all apologies and only resend the position if it is out of the leniency.

PR #748, opened 2017-11-26
Fix encryption in pipeline for Glowstone

momothereal:
It looks like encryption was only working for the Spigot pipeline, so I added a check for Glowstone and replicated the correct order for GS's channel pipeline.

~After this fix, a client receiving a chat packet (0x0F), like the join message, will crash. I'm not sure if this is caused by this particular fix or by another issue in ProtocolSupport.~ This is fixed, I swapped the encryption/decryption modes by accident.

### DIFF
--- a/src/protocolsupport/zplatform/impl/glowstone/GlowStoneMiscUtils.java
+++ b/src/protocolsupport/zplatform/impl/glowstone/GlowStoneMiscUtils.java
@@ -23,8 +23,12 @@ import net.glowstone.net.pipeline.MessageHandler;
 import net.glowstone.net.protocol.ProtocolType;
 import net.glowstone.util.GlowServerIcon;
 import protocolsupport.api.utils.NetworkState;
+import protocolsupport.protocol.pipeline.ChannelHandlers;
 import protocolsupport.protocol.pipeline.IPacketPrepender;
 import protocolsupport.protocol.pipeline.IPacketSplitter;
+import protocolsupport.protocol.pipeline.common.PacketDecrypter;
+import protocolsupport.protocol.pipeline.common.PacketEncrypter;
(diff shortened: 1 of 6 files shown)


Shevchik:
I think enabling encryption should be moved to PlatformMiscUtils instead

src/protocolsupport/zplatform/impl/glowstone/GlowStoneMiscUtils.java
```diff
@@ -200,6 +207,12 @@ public void enableCompression(ChannelPipeline pipeline, int compressionThreshold
 		pipeline.addAfter(GlowStoneChannelHandlers.FRAMING, "compression", new CompressionHandler(compressionThreshold));
 	}
 
+	@Override
+	public void enableEncryption(ChannelPipeline pipeline, SecretKey key, boolean fullEncryption) {
+		pipeline.addBefore(GlowStoneChannelHandlers.FRAMING, ChannelHandlers.ENCRYPT, new PacketEncrypter(MinecraftEncryption.getCipher(Cipher.ENCRYPT_MODE, key)));
+		pipeline.addBefore(ChannelHandlers.ENCRYPT, ChannelHandlers.DECRYPT, new PacketDecrypter(MinecraftEncryption.getCipher(Cipher.DECRYPT_MODE, key)));
```
Shevchik: You forgot to check if encrpytion is full

momothereal: Is full encryption a cross-platform thing? If so, what's the pipeline handler ordering for it?

Shevchik: Full encryption means that both encryption handlers are present. Not full means that only decryption handler should be present.

PR #750, opened 2017-11-29
Possible solution for stuttering objects.

7kasper:
Is this a crime?
----
The stuttering arrows and other objects are quite annoying, especially in PVP.
Since PE doesn't do motion (except for ItemEntities) the only solution is to send positions more often.
The server does this already for most mobs, but objects usually have a pesky updateInterval to save performance. So, why not just send movement updates always when there is a good velocity? 😉 

Perhaps we can optimise it a bit more, or maybe not 🤷‍♂️


7kasper:
Bumb :p

7kasper:
@Shevchik Do you think we can give this a try? We can always revert it if it affects performance significantly.

PR #755, opened 2017-12-05
Fix UnsupportedOperationException under Paper

MrPowerGamerBR:
📰 📱 

Same thing as https://github.com/ProtocolSupport/ProtocolSupport/pull/723 however now it doesn't use Paper as a dependency, it can be compiled against Spigot without any issues. (I tried "optimizing" the Reflection stuff so it won't have an big performance penalty)

The reason ProtocolSupportPE's throws UnsupportedOperationException under Paper is because `trackedPlayers` is actually `trackedPlayerMap`'s keySet(), so, instead of editing `trackedPlayers` we need to edit the `trackedPlayerMap`.

We also need to set the `tracker` in the NMS Entity.

To avoid code duplication, `PaperEntityTrackerEntry` extends `SpigotEntityTrackerEntry`, that's why some of the fields/methods are now protected.

While I didn't do extensive tests, it seems there isn't any issues when spawning entities/players.


src/protocolsupport/zplatform/impl/paper/entitytracker/PaperEntityTrackerEntry.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Map;
+
+public class PaperEntityTrackerEntry extends SpigotEntityTrackerEntry {
+	private static Field TRACKED_PLAYER_MAP_FIELD;
+	private static Field TRACKER_ENTITY_FIELD;
+	private Map<EntityPlayer, Boolean> trackedPlayerMap;
+
+	static {
+		try {
+			TRACKED_PLAYER_MAP_FIELD = EntityTrackerEntry.class.getDeclaredField("trackedPlayerMap");
```
7kasper: `ReflectionUtils.getField(EntityTrackerEntry.class, "trackedPlayerMap");`
setAccessible is then already done.

src/protocolsupport/zplatform/impl/paper/entitytracker/PaperEntityTrackerEntry.java
```diff
@@ -0,0 +1,60 @@
...
+		try {
+			trackedPlayerMap = (Map<EntityPlayer, Boolean>) TRACKED_PLAYER_MAP_FIELD.get(this);
+		} catch (IllegalAccessException e) {
+			e.printStackTrace();
+		}
+	}
+
+	@Override
+	protected void addTrackedPlayer(EntityPlayer player) {
+		this.trackedPlayerMap.put(player, true);
```
7kasper: You are adding to the trackedPlayerMap for sure, and adding to the tracked PE and Default players will make some packets to be send. BUT... don't methods such as `void clear(final EntityPlayer entityplayer)` or even the normal 'public void broadcast(final Packet<?> packet)' depend on the trackedPlayers set? As far as I can see, this set will always be empty.

MrPowerGamerBR: hmmmm yeah, I think I need to test it then.

I just need to figure out an way to see if it would really break or not.

7kasper: @MrPowerGamerBR have you figured it out yet? If Shevchik is okay with this, I am too :P

MrPowerGamerBR: @7kasper yeah, don't you remember the pig sacrifice I did? rip smol piggy

You asked me to test if the entity metadata was still working after this change, and it was working, I tried igniting a pig and it did work.

7kasper: Yes I do remember that.

MrPowerGamerBR:
After months of testing this PR on my test server, it seems it works fine. 🎉

However I'm not sure if I should change the "check if it is Paper" condition to get if any Paper classes are available instead of checking the Bukkit version. (the first way would make it work in any Paper forks)

Shevchik:
Paper shouldn't have it's own platform identifier

MrPowerGamerBR:
@Shevchik done, I removed the ServerPlatformIdentifier.


I'm still wondering if I should replace the "check if it is Paper" condition to a `Class.forName("...")` instead of checking the version name... I can't do what ProtocolSupport does right now because we aren't compiling against Paper.

PR #759, opened 2017-12-26
Use abs time in <= 1.5.2

EntryPointKR:
Flicker problem is in 1.5.2 too.
and 1.6.4 or later is not


Shevchik:
Indeed doDayNightCycle gamerule was only added in 1.6, thanks for noticing it.

Issue #739, opened 2017-11-02
Implement Paper's PlayerHandshakeEvent


lucko:
Paper implements a `PlayerHandshakeEvent` in these patches:

https://github.com/PaperMC/Paper/blob/master/Spigot-API-Patches/0034-Add-handshake-event-to-allow-plugins-to-handle-clien.patch

https://github.com/PaperMC/Paper/blob/master/Spigot-Server-Patches/0112-Add-handshake-event-to-allow-plugins-to-handle-clien.patch

It allows plugins to handle handshaking logic themselves, without the need to perform server-side modifications. This allows anyone running a custom proxy solution (BungeeCord, LilyPad, etc - forks included) to handle authentication themselves, without performing modifications to the server.

When running ProtocolSupport, this event is never called, as PS uses it's own implementation.

I understand it's possible to achieve a similar effect by registering a `SpoofedData` parser in the `SpoofedDataParser` class, but I'd be nice if PS could support Paper's event too. :)

Among other things, the event allows you to define the failure reason, which is handy, as the default one used by PS doesn't apply to all situations.

> Ip forwarding is enabled but spoofed data can't be decoded or is missing

Just for reference, this is the plugin which made me notice this issue: https://github.com/lucko/BungeeGuard

I might attempt a PR if I get some time in the next few weeks (and if you indicate you'd accept this change). Only slight issue is it means somehow depending on the Paper API to gain access to the event, or just copying the class into ProtocolSupport. I understand Glowstone also forks off Paper, so I don't think this would be an issue?

Shevchik:
The paper dependency should be a soft depend.

PS has ConnectionHandshakeEvent, but it is broken right now, should fix that.

PR #761, opened 2017-12-28
Add support for Paper's PlayerHandshakeEvent

lucko:
Finally got some free time to PR a fix for the issue I submitted here: #739

Let me know if there are any changes needed. :)

### DIFF
--- a/src/protocolsupport/protocol/utils/spoofedata/PaperSpoofedDataParser.java
+++ b/src/protocolsupport/protocol/utils/spoofedata/PaperSpoofedDataParser.java
@@ -1,0 +1,34 @@
+package protocolsupport.protocol.utils.spoofedata;
+
+import com.destroystokyo.paper.event.player.PlayerHandshakeEvent;
+
+import org.bukkit.Bukkit;
+
+import protocolsupport.api.events.PlayerPropertiesResolveEvent.ProfileProperty;
+import protocolsupport.utils.Utils;
+
+import java.util.UUID;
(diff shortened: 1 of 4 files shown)


src/protocolsupport/protocol/utils/spoofedata/PaperSpoofedDataParser.java
```diff
@@ -0,0 +1,34 @@
+package protocolsupport.protocol.utils.spoofedata;
+
+import com.destroystokyo.paper.event.player.PlayerHandshakeEvent;
```
7kasper: 🤔

src/protocolsupport/protocol/utils/spoofedata/SpoofedDataParser.java
```diff
@@ -9,17 +9,33 @@
...
 			} catch (Exception e) {
-				return null;
+				// ignore
 			}
-		}).filter(res -> res != null).findAny().orElse(null);
+		}
+		return null;
+	}
+
+	private static boolean isPaperHandshakeEvent() {
```
7kasper: We might combine this with @MrPowerGamerBR's mcpe paper commit. Both PR's should maybe rely on paper detection at start and not searching class files in runtime to see if we're running paper or not.

lucko: It's more than just "Paper" detection.

Older versions didn't provide the event.

PR #765, opened 2018-01-27
add suport for 1.2.10.1

Icey-girl:


Icey-girl:
i know main protocol of the beta and the normal one why dont add it?

PR #779, opened 2018-02-24
[PSPE] Possible fix for block break animations.

7kasper:
Sends block break animations to PE player if it is the one mining the block too. This way the faker can catch the block break animation.


Shevchik:
spaces -> tabs

PR #781, opened 2018-02-24
BlockUpdate, why not?

7kasper:


src/protocolsupport/listeners/InternalPluginMessageRequest.java
```diff
@@ -71,7 +76,7 @@ public ChunkUpdateRequest(int chunkX, int chunkZ) {
 			this.chunkX = chunkX;
 			this.chunkZ = chunkZ;
 		}
-
+
```
Shevchik: !

src/protocolsupport/listeners/InternalPluginMessageRequest.java
```diff
@@ -108,6 +141,8 @@ public static void receivePluginMessageRequest(Connection connection, PluginMess
 
 	@Override
 	public void onPluginMessageReceived(String tag, Player player, byte[] data) {
+		Connection connection = ProtocolSupportAPI.getConnection(player);
+		if (connection == null) { return; }
```
Shevchik: if (connection == null) {
return;
}

PR #784, opened 2018-03-02
Set java.awt.headless to true to disable app window on macos.

magicus:
When running on macosx, an app window (icon + menues) appear from Spigot when loading the PS plugin. This is due to the fact that a few classes in PS utilise AWT, such as java.awt.Color. As soon as you use any AWT, it starts loading a lot of GUI stuff. Most of it can be disabled by setting the property java.awt.headless to true.

I'm not sure this is a good placement of the code, but I found no good init() function to put it in.


7kasper:
If this is done it should be in the main branch and perhaps in the constructor.

Shevchik:
^

PR #786, opened 2018-03-04
Fix issue with spigot jar constantly re-downloading.

magicus:
Special handling to check if file exists. Since we use specific version, and not a "latest" from a CI system, this should be fine.

Also refactored to get file name with versions in just one place.


Shevchik:
Fixed by commit #bae48c2

Issue #787, opened 2018-03-05
[Feature] Maven repo


DiscoverSquishy:
Hey. I wanna start a project but I want to use this and add it into my pom.xml file although I need a repository it would be great if the team could add one unless I’m blind and there is one already.

DaMatrix:
if the jars were to be downloaded automatically then you could use

https://jitpack.io/#ProtocolSupport/ProtocolSupport

but it doesn't work currently

DiscoverSquishy:
Bump

games647:
JitPack has an older version in their repository, but a lot have changed since. 

```xml
        <dependency>
            <groupId>com.github.ProtocolSupport</groupId>
            <artifactId>ProtocolSupport</artifactId>
            <!--4.25.dev-->
            <version>a4f060dc46</version>
            <scope>provided</scope>
        </dependency>
```

Default Gradle projects install their jar into `build/libs/project-name-version.jar`, but this project stores them into `target/project-name.jar`. Therefore JitPack cannot find it. Either Shevchik should install a maven repository server on the build server or we add a copy/symlink to the expected folder, so JitPack can find it.

Shevchik:
I guess we can copy jar from build libs to target folder for now.

Shevchik:
You can now use jitpack for that.

PR #796, opened 2018-03-19
Allow JitPack to find the output jar by copying it

games647:
(Related #787)

This builds the jar to the normal Gradle output directory (with normal project name and version suffix) and copies it after it to target/ProtocolSupport.jar (without version suffix).

JitPack builds:
[Build log](https://jitpack.io/com/github/games647/ProtocolSupport/jitpack-7c18ea65ca-1/build.log)
[Output jar](https://jitpack.io/com/github/games647/ProtocolSupport/jitpack-7c18ea65ca-1/ProtocolSupport-jitpack-7c18ea65ca-1.jar)

Other project can then use something like this.
```xml
	<repositories>
		<repository>
		    <id>jitpack.io</id>
		    <url>https://jitpack.io</url>
		</repository>
	</repositories>
```

```xml
	<dependency>
	    <groupId>com.github.games647</groupId>
	    <artifactId>ProtocolSupport</artifactId>
            <!-- 4.29-dev -->
	    <version>jitpack-7c18ea65ca-1</version>
	</dependency>
```

For versioning you could tag each new released version using Git. Then we could replace the commit hash with the actual version representation like this below.

```xml
	<dependency>
	    <groupId>com.github.games647</groupId>
	    <artifactId>ProtocolSupport</artifactId>
	    <version>4.29-dev</version>
	</dependency>
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,10 +31,7 @@ tasks.withType(JavaCompile) {
 
 group 'protocolsupport'
 version '4.29-dev'
-sourceCompatibility = 1.8
-
-
-import java.text.MessageFormat
+sourceCompatibility = JavaVersion.VERSION_1_8
 
 import java.text.MessageFormat
 
@@ -144,6 +141,15 @@ task generateInfo(type: DefaultTask) {doLast{
 	new FileOutputStream(buildInfoFile).withCloseable({ properties.store(it, "Build info") })
 }}
 
+task copyFinalJarToTarget(type: Copy) {
+	// JitPack searches for the output jar at the standard Gradle output directory (jar.archivePath)
+	// By copying it from there to our target destination JitPack can archive it in a Maven repository
+	from jar.archivePath.getPath()
+	into 'target'
+
+	//remove version suffix
+	rename (jar.archiveName, jar.baseName + '.jar')
+}
 
 shadowJar {
 	doFirst {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -144,6 +141,16 @@ task generateInfo(type: DefaultTask) {doLast{
...
 }}
 
+// copy jar from the standard Gradle output (build/libs) to target/
+// This allows JitPack to find the output jar and archive it
+task copyFinalJarToTarget(type: Copy) {
+	from buildDir.getName() + '/' + libsDirName + '/' + project.name + '-' + project.version + '.jar'
+	into 'target'
+
+	//remove version suffix
+	rename (project.name + '-' + project.version + '.jar',
```
Shevchik: 1 line pls

build.gradle
```diff
@@ -144,6 +141,15 @@ task generateInfo(type: DefaultTask) {doLast{
 	new FileOutputStream(buildInfoFile).withCloseable({ properties.store(it, "Build info") })
 }}
 
+// copy jar from the standard Gradle output (build/libs) to target/
+// This allows JitPack to find the output jar and archive it
+task copyFinalJarToTarget(type: Copy) {
```
7kasper: copyFinalJarToLibs is alot more clear. Also you probably don't need those comments.

build.gradle
```diff
@@ -144,6 +141,15 @@ task generateInfo(type: DefaultTask) {doLast{
...
+	from buildDir.getName() + '/' + libsDirName + '/' + project.name + '-' + project.version + '.jar'
+	into 'target'
+
+	//remove version suffix
+	rename (project.name + '-' + project.version + '.jar', project.name + '.jar')
```
7kasper: I see a duplicate of archivename, maybe reuse the variable in here as well.

build.gradle
games647: ? It copies from `build/libs` to `target/` and not in reverse order.

I would keep line 145, because it states why we introduced this change.

7kasper: @games647 Yeah I saw, edited before I read your answer 😄

7kasper:
👍

PR #814, opened 2018-05-02
Sound name size make to <= 32

EntryPointKR:
If name size more than 32, would be crash in 1.4.7 & 1.5.2. but 1.6 or higher isn't.
Is this correct way for solve that?


Shevchik:
Utils::clampString

PR #815, opened 2018-05-02
Add hit sound remap entry in <= 1.6

EntryPointKR:
Doesn't play hurt sound in <= 1.6


src/protocolsupport/protocol/typeremapper/sound/SoundRemapper.java
```diff
@@ -124,6 +124,7 @@
 			registerRemapEntry("entity.hostile.death", "damage.hit", ProtocolVersionsHelper.BEFORE_1_7);
 			registerRemapEntry("entity.generic.hurt", "damage.hit", ProtocolVersionsHelper.BEFORE_1_7);
 			registerRemapEntry("entity.player.death", "damage.hit", ProtocolVersionsHelper.BEFORE_1_7);
+			registerRemapEntry("entity.player.hurt", "damage.hit", ProtocolVersionsHelper.BEFORE_1_6);
```
7kasper: Are you sure 1.7 doesnt need this remap too? From the remap above it seems like it.

PR #816, opened 2018-05-03
Particle count make to <= 1

EntryPointKR:
Can see particle even if particle count is 0 in 1.7 or higher. but <= 1.6 isn't
In 1.7 or higher, particle count 0 work like 1. if less than 0, not displayed.

Should I check version <= 1.6 and count is 0? or just keep this code?


EntryPointKR:
Oh, 0 is separate operation.

mcp

So I will edit to check version <= 1.6

PR #819, opened 2018-05-12
Fix minecart position issue

EntryPointKR:
Is this all of the minecart problem, no others?
I could reproduce only problem that if rejoin, minecart passenger looks flying.
So this PR for that


src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_4_5_6_7/AbstractSpawnObject.java
```diff
@@ -0,0 +1,85 @@
...
+				y += 16;
+				break;
+			}
+			case ARROW: {
+				objectdata--;
+				break;
+			}
+			case MINECART: {
+				if (version.isAfterOrEq(ProtocolVersion.MINECRAFT_1_6_1)) {
+					y += 16;
```
EntryPointKR: For minecart location problem

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_4_5/VehiclePassengers.java
```diff
@@ -24,10 +29,20 @@
...
 		}
+		NetworkEntity vehicle = cache.getWatchedEntityCache().getWatchedEntity(vehicleId);
+		NetworkEntity passenger = cache.getWatchedEntityCache().getWatchedEntity(passengerId);
+		if (vehicle != null && passenger != null && passengersIds.length != 0) {
+			if (passengers.contains(passengerId)) {
+				return false;
+			} else {
+				passengers.add(passengerId);
+			}
+		}
```
EntryPointKR: For passenger position problem

EntryPointKR:
> your player is rapidly switching between the "Sitting in minecart" and "Standing looking at minecart" positions. Because of this sometimes very rarely your player (the one in the minecart) gets teleported 2-3 blocks above the minecart and the game softlocks, as in everything is frozen except the chat (for that player).

**Couldn't reproduce but when non-passenger rejoin, passenger looks flying(minecart above 2~3blocks). that fixed**

> Commands that kill the frozen player make the bone-breaking sound and health flashes however it does not seem to decrease or have any other effect. Upon logging out and back in you get the "Game over" screen because the damage was actually applied and when you click respawn everything looks OK at first however other weird glitches arise such as you can't see projectiles in mid flight, falling sand disappears and re appears at bottom etc.

**Couldn't reproduce. but I found bug that can't respawn**

> For other players, the player in the minecart is rapidly appearing riding the minecart and switching to a position in which he is standing close to the minecart and looking at it (constantly). This can be verified by building a small minecart loop and running a cart inside it. The player riding the minecart switches between looping around and being at the center and spinning around looking at the minecart.

**Couldn't reproduce**

> If you stay in the minecart long enough (5 or 10 minutes) and press "Sneak" you might get a "End of stream" error

**Will be edit**

EntryPointKR:
I'll pull request again for explain

PR #820, opened 2018-05-13
Fix minecart issue

EntryPointKR:
> your player is rapidly switching between the "Sitting in minecart" and "Standing looking at minecart" positions. Because of this sometimes very rarely your player (the one in the minecart) gets teleported 2-3 blocks above the minecart and the game softlocks, as in everything is frozen except the chat (for that player).

**Couldn't reproduce but when non-passenger rejoin, passenger looks flying(minecart above 2~3blocks). This fixed**

> Commands that kill the frozen player make the bone-breaking sound and health flashes however it does not seem to decrease or have any other effect. Upon logging out and back in you get the "Game over" screen because the damage was actually applied and when you click respawn everything looks OK at first however other weird glitches arise such as you can't see projectiles in mid flight, falling sand disappears and re appears at bottom etc.

**Couldn't reproduce. but I found bug that can't respawn**

> For other players, the player in the minecart is rapidly appearing riding the minecart and switching to a position in which he is standing close to the minecart and looking at it (constantly). This can be verified by building a small minecart loop and running a cart inside it. The player riding the minecart switches between looping around and being at the center and spinning around looking at the minecart.

**Couldn't reproduce**

> If you stay in the minecart long enough (5 or 10 minutes) and press "Sneak" you might get a "End of stream" error

**Couldn't reproduce. tested 10m and 30m**

Did I miss anything?

### DIFF
--- a/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_4_5/VehiclePassengers.java
+++ b/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_4_5/VehiclePassengers.java
@@ -4,12 +4,14 @@ import it.unimi.dsi.fastutil.ints.Int2IntOpenHashMap;
 import protocolsupport.protocol.packet.ClientBoundPacket;
 import protocolsupport.protocol.packet.middle.clientbound.play.MiddleVehiclePassengers;
 import protocolsupport.protocol.packet.middleimpl.ClientBoundPacketData;
+import protocolsupport.protocol.storage.netcache.WatchedEntityCache;
 import protocolsupport.utils.recyclable.RecyclableCollection;
 import protocolsupport.utils.recyclable.RecyclableSingletonList;
 
 public class VehiclePassengers extends MiddleVehiclePassengers {
 
 	protected final Int2IntOpenHashMap vehiclePassenger = new Int2IntOpenHashMap();
+	protected final Int2IntOpenHashMap passengerToVehicle = new Int2IntOpenHashMap();
 	protected int passengerId;
 
 	@Override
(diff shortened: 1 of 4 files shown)


src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_4_5/VehiclePassengers.java
```diff
@@ -24,10 +26,20 @@
...
 		}
+		WatchedEntityCache entityCache = cache.getWatchedEntityCache();
+		if (passengersIds.length != 0 && entityCache.hasWatchedEntity(vehicleId) && entityCache.hasWatchedEntity(passengerId)) {
+			int prevVehicleId = passengerToVehicle.getOrDefault(passengerId, -1);
+			if (entityCache.hasWatchedEntity(prevVehicleId)) {
+				return false;
+			} else {
+				passengerToVehicle.put(passengerId, vehicleId);
+			}
+		}
```
EntryPointKR: This for problem that if non-passenger rejoin, minecart passenger looks flying above 2~3 blocks.
Cause 1.12.2 bukkit sends PacketPlayOutMount(VehiclePassenger) 3times. 1.5.2 isn't

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_4_5_6/SpawnObject.java
```diff
@@ -59,6 +59,12 @@
 				objectdata--;
 				break;
 			}
+			case MINECART: {
+				if (version.isAfterOrEq(ProtocolVersion.MINECRAFT_1_6_1)) {
+					y += 16;
+				}
+				break;
+			}
```
EntryPointKR: This for minecart y location problem in 1.6~1.7.
In 1.6~1.7, minecart passenger looks down position likes 1.5.2.
So I added y 16(0.5)

EntryPointKR: And 1.5.2 need it too.
So if statement will be remove.

Shevchik: Remove the if statement then.

EntryPointKR: Ok

PR #821, opened 2018-05-18
Fix position desync bug

EntryPointKR:
The cause of this problem is that difference as much as the remainder when dividing.


Shevchik:
That explains why we have this issue, finally we can fix that. I will use other method to calculate that, because adding the remainder looks more like a workaround.

PR #822, opened 2018-05-19
Update protocol and version number [pspe]

XxKurumi-TokisakixX:
I do more work later but I need to help my dad first


7kasper:
Thanks but it needed to be in mcpenew. I already changed.

XxKurumi-TokisakixX:
@7kasper alright I will look if I can help again

PR #838, opened 2018-06-18
Give correct error message if a too old PE client tries to connect

magicus:
(instead of claiming that the client is too new)


MrPowerGamerBR:
Doesn't Minecraft has a "outdated client" message (Spigot can customize that message, so I guess it does also exist in vanilla MC)? If it has, you should use it for consistency.

7kasper:
I'll cherry-pick this another time.

PR #839, opened 2018-06-20
Lazily load translations

astei:
Right now, we load all translations into memory. This is wasteful, especially since most locales we load will never get used (I observed ~52MB being used). This is important in our memory constrained environment.

This changes `I18NData` so that all locales are lazily loaded, with the exception of `en_us` (which will always be loaded).

More savings are potentially possible, but I did not investigate as this was considered to be good enough for most users.


Shevchik:
The problem with lazy load like that is that it will be triggered on netty io thread. Sudden ping spike ensures.

Shevchik:
Maybe instead of lazy load it all should be tied to enabled/disabled protocol versions?

7kasper:
@Shevchik I understand we rather not do any kind of loading on netty thread, but if we load the lang at async login a slight delay might be a compromise for the high memory usage.

astei:
I'd be happy to change this so that locales are loaded on login and on settings change packets.

astei:
I got a little bit of time to work on this. I've changed this so that the locale is loaded asynchronously when a `Client Settings` packet is sent. I've implemented this for Spigot, but I'm looking where I can implement this for Glowstone. I haven't found it just yet, but I'll pick this up when I get the time.

Shevchik:
While this exact pull request can't be accepted, I will still implement something similar to this, but using PlayerLocaleChangeEvent. It will asynchronously load the locale from default. Also new API methods will be added to TranslationAPI that will allow quering loaded default locale, unload, and load them (so you can write a plugin that preloads locales beside the default en_us one).

PR #849, opened 2018-07-11
Fix updates on furnace interface

magicus:
Fixes cooking (which was twice the speed) and fuel consumption (which was not shown at all).


7kasper:
I still think it changed in pe sometime. I swear it was working properly once. Well anyway good job!

magicus:
Yeah, it can very well have changed in PE. Mojang seems to like to modify things willy-nilly.

But it has been broken at least since 1.2.10, which has been my "stable" version.

PR #850, opened 2018-07-11
GSON fails to match PEEntityData fields from entitydata.json

magicus:
The names of the fields in PEEntityData must match exactly the format in entitydata.json, including case.

Without this fix, the information in entitydata.json is basically lost.

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
+++ b/src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
@@ -4,6 +4,7 @@ import java.io.BufferedReader;
 import java.util.EnumMap;
 import java.util.Map;
 
+import com.google.gson.annotations.SerializedName;
 import org.bukkit.enchantments.Enchantment;
 import org.bukkit.util.Vector;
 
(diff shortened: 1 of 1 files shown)


magicus:
This fixes at least my observations in https://github.com/ProtocolSupport/ProtocolSupportPocketStuff/issues/80, and possibly multiple other issues as well (since no data from entitydata.json, except InventoryFilter, has been used).

src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
```diff
@@ -406,21 +406,23 @@ public static PEEntityData getEntityData(NetworkEntityType type) {
 	}
 
 	public static class PEEntityData {
-		private BoundingBox boundingBox;
-		private Offset offset;
-		private RiderInfo riderInfo;
+		// Note that these fields must have non-standard naming to match
+		// the format in entitydata.json
+		private BoundingBox BoundingBox;
```
MrPowerGamerBR: Please use Gson's SerializedName annotation on every field, instead of changing the field name, to keep consistency between classes

`@SerializedName("BoundingBox")`
`private BoundingBox boundingBox`

etc etc etc

PR #851, opened 2018-07-12
update remaps for 1.5.0 [pspe]

Wesley-VanNeck:


magicus:
This solves the remap issues with colored stuff, like glass block. @7kasper, perhaps you can accept this?

PR #853, opened 2018-07-15
Fix bounding box for ender dragon

magicus:
Also print warning if bounding box is missing from json for future problems like this.

Fixes PSPS#77.

### DIFF
--- a/resources/resources/pe/entitydata.json
+++ b/resources/resources/pe/entitydata.json
@@ -815,6 +815,12 @@
 			"height": 0.31
 		}
 	},
+	"ENDER_DRAGON": {
+		"BoundingBox": {
+			"width": 13,
+			"height": 4
+		}
+	},
 	"EVOCATOR_FANGS": {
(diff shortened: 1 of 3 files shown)


magicus:
```
assets/behavior_packs/vanilla/entities/ender_dragon.json:      "minecraft:collision_box": {
assets/behavior_packs/vanilla/entities/ender_dragon.json-        "width": 13,
assets/behavior_packs/vanilla/entities/ender_dragon.json-        "height": 4
assets/behavior_packs/vanilla/entities/ender_dragon.json-      },
```
from the Android APK.

magicus:
I can't really say for sure that it's correct. I tried hitting the dragon with the sword, nothing happened. (Maybe this is supposed to be the case; I've never really played against the dragon for real). I know the boss bar is not working so it's hard to tell if something happened.

But at the very least we're not crashing anymore, so of there are more issues with the dragon left we can debug them further on.

PR #858, opened 2018-07-23
Move block ID remaps from external PHP script into core Java code.

magicus:
I have verified programmatically that the resulting remaps are identical to what was produced by the old code. (Verification code removed before final commit.)


magicus:
@Wesley-VanNeck This is a rewrite to have all remapping logic in the Java code base, instead of having it in a separate pop script.

src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
```diff
@@ -137,9 +140,177 @@ public static NetworkEntityType getLivingTypeFromPeNetworkId(int networkId) {
 	}
 
 	public static final ArrayBasedIdRemappingTable BLOCK_ID = new ArrayBasedIdRemappingTable(MinecraftData.BLOCK_ID_MAX * MinecraftData.BLOCK_DATA_MAX);
+
+	private static HashMap<Integer, ArrayList<Integer>> blockIdRemaps = new HashMap<>();
```
yannicklamprecht: Follow LSP.

src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
```diff
@@ -137,9 +140,177 @@ public static NetworkEntityType getLivingTypeFromPeNetworkId(int networkId) {
...
+	private static HashMap<Integer, ArrayList<Integer>> blockStateRemaps = new HashMap<>();
```
yannicklamprecht: LSP

src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
```diff
@@ -137,9 +140,177 @@ public static NetworkEntityType getLivingTypeFromPeNetworkId(int networkId) {
...
+		blocksJson.forEach(entry -> {
+			// For most blocks, map PC blockState to PE runtime ID. For some blocks,
+			// the PE id/blockState is different from the PC id/blockState. These
+			// are encoded in our remaps.
+			int runtimeID = entry.getAsJsonObject().get("runtimeID").getAsInt();
+			int id = entry.getAsJsonObject().get("id").getAsInt(); // PE id
+			int data = entry.getAsJsonObject().get("data").getAsInt(); // PE data
+
+			int blockState = (id << 4) | data;
+			ArrayList<Integer> remappedIds = blockIdRemaps.get(id);
```
yannicklamprecht: LSP

src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
```diff
@@ -137,9 +140,177 @@ public static NetworkEntityType getLivingTypeFromPeNetworkId(int networkId) {
...
+			ArrayList<Integer> remappedBlockStates = blockStateRemaps.get(blockState);
```
yannicklamprecht: LSP

7kasper:
Thank you. I will be porting it myself as to include the new changes for 1.13.

yannicklamprecht:
By the way it’s JSON configuration file. I see only advantages using the data driven way.

src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
Shevchik: We don't require following LSP actually, but this field is actually missing an final modifier, which is more important IMO.

yannicklamprecht:
@7kasper with LSP or with data driven approach?

LSP -> Liskov Substitution Principle -> develop against abstraction instead of implementation. 

data driven approach -> instead of hard coding everything you can add configuration to improve adaptation speed.

PR #860, opened 2018-07-24
Add proper remaps for entity status codes.

magicus:
Fixes missing love/heart particles, totem of undying animation, etc. Will close PSPE#30.

### DIFF
--- a/src/protocolsupport/ProtocolSupport.java
+++ b/src/protocolsupport/ProtocolSupport.java
@@ -173,6 +173,10 @@ public class ProtocolSupport extends JavaPlugin {
 		}
 	}
 
+	public static void logTrace(String message) {
+		ProtocolSupport.getInstance().getLogger().fine(message);
+	}
+
 	public static void logInfo(String message) {
 		ProtocolSupport.getInstance().getLogger().info(message);
 	}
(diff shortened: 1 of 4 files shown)


caoli5288:
👍

7kasper:
Good work. Now the null mappings are a good idea, but I have something better. When I wrote the networkentitytype I created a helper function that can check supertype as well. Sort of like an instanceof operator. This way you can do remaps to Entity (which applies to all entities) or perhaps tameable (so you dont have to remap for wolf and ocelot twice for which they have in common. Check the `entityType.isOfType(entityType);` function. To store the remaps use an array or int2objecthashmap. The int used as key is the pcStatus as value you can use something like: `ImmutablePair<Integer, NetworkEntityType>` You get the remaps using something like: `map.get(pcStatus).stream().filter(pair -> entityType.isOfType(pair.getRight()).findFirst().orElse(ImmutablePair.of(-1, NetworkEntityType.Entity).getLeft();`

I dont know if I am making it too complicated RN.. If you want give it a try otherwise Iĺl port it over myself when I get time.

magicus:
Adding support for more remaps might show better how the framework needs improving. (For a start, PEDataValues should be a singleton. :))

magicus:
Adding support for more remaps might show better how the framework needs improving. (For a start, PEDataValues should be a singleton. :))

7kasper:
This is still very much relevant after 1.13 yes. Unlike your other pr but ill look at that tomorrow.

Missing remaps can be done quite easily but we might need to digg through the pe binary for that.

I am thinking of more complicated framework but maybe better for remaps. If you like ill accept this tomorrow with some minor changes, but I might break it so it will need some testing.

magicus:
Do whatever changes you feel are needed. It'll probably be some days before I'll be able to test anything, though.

PR #873, opened 2018-08-16
Merge fixinv to new

connor557:


connor557:
Gotta switch base and compare branches

PR #894, opened 2018-09-10
Add missing PE entity IDs for new mobs (fish varietes, etc).

magicus:
Unfortunately, I've not yet been able to determine correct IDs for all types. But without any mapping at all, we crash with NPE when the entity spawn. :(


7kasper:
Great, thanks :)

PR #904, opened 2018-09-23
Fix legacy block states for melon/pumpkin stems and brick stairs.

magicus:
Also improve handling of carved/plain pumpkins (but directions are still wrong).


7kasper:
Awesome! Thank you.

magicus:
@7kasper No, thank you! This was if not mostly your work, at least co-written by you. I could not have fixed this without your help.

PR #906, opened 2018-10-01
Update note_block sounds mappings

zedwick:
This just adds note_block sound mappings for 1.9 through 1.12.2, and updates the mappings for pre-1.9.

This is tested in 1.12.2, but as I understand it should work for previous versions (I figured I may as well update the pre-1.9 mappings whilst there, as I _think_ they needed updating. Please correct me if I'm wrong there.)

I wanted to update these mappings specifically for a song played via the /playsound command in an mcfunction, but I found out this doesn't work for the playsound command. This appears to only work when a note block produces the sound. Any guidance for how to go about adding support for /playsound would be very welcome!


7kasper:
Thanks!

PR #908, opened 2018-10-06
Updated mappings for 1.13 flattenings.

zedwick:
This should update all remaining mappings for sounds changed in the great flattening of 1.13.

Due to the scope I've not tested all the sounds individually, but I believe they should all be correct and working.


7kasper:
Ok great! Thanks again 😄

Issue #889, opened 2018-09-09
clicking on the crafting table kicks you in 1.7.10


YasunaMC:
![20180909_134906](https://user-images.githubusercontent.com/25864932/45264101-578f2500-b437-11e8-8ac6-f1d4934fe499.jpg)

Shevchik:
You mean trying to open that?

WesleyVanNeck:
@Shevchik yes

PR #909, opened 2018-10-06
Fix crafting table open error

EntryPointKR:
When open the crafting table, current PS will send a inv title exactly "Unknown translation key: block.minecraft.crafting_table.name" to 1.4-1.7
Then 1.4~1.7 will validate the title size, crashes.
Refer issue: #889

### DIFF
--- a/src/protocolsupport/protocol/packet/middle/clientbound/play/MiddleInventoryOpen.java
+++ b/src/protocolsupport/protocol/packet/middle/clientbound/play/MiddleInventoryOpen.java
@@ -47,18 +47,6 @@ public abstract class MiddleInventoryOpen extends ClientBoundMiddlePacket {
 	}
 
 	protected String getLegacyTitle() {
-		switch (type) {
-			// Title users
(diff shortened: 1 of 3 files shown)


Shevchik:
Translation files needs to be updated to actually translate that key instead of hacking around. Name clamping is still needed tho.

EntryPointKR:
Thanks for review

EntryPointKR:
Oh wait. Does "Only clamp the name" mean like that?
The workbench title is not required in 1.4~1.7 so I replaced to empty string

src/protocolsupport/protocol/packet/middle/clientbound/play/MiddleInventoryOpen.java
```diff
@@ -45,4 +46,7 @@ public boolean postFromServerRead() {
 		}
 	}
 
+	protected String getLegacyTitle() {
```
Shevchik: inline to callers

Issue #905, opened 2018-09-24
Maps Bug (BannerBoard)


EazyFTW:
Is it possible for you to fix maps with BannerBoard. At the moment when you join in a version other than 1.13.x, the maps show nothing (Just an empty item frame, but bigger like a transparent map). I contacted their support and they said it's an issue with ProtocolSupport. 

Thanks,
- EazyFTW.

Shevchik:
Related to item durability not being copied from nbt to legacy data

mibby:
@Shevchik Do you have any estimate when this can be fixed? I also rely heavily on having map data properly sent to older client versions.

7kasper:
Should be fixed now! 😄

TheDumbledodo:
There is this bug with bannerboard that when there are bungee servers the images from (lobby) pastes on like (Bedwars) if you make 2 maps it won't be bugged but if you reload the server the maps will be again ripped I got Protocolib too
anyone could help? I've had this plugin for a goddamn month

Issue #900, opened 2018-09-18
Tools/Weapons/Armor do not show damage (1.12.2 on 1.13.1)


Brokkonaut:
When playing in 1.12.2 on an 1.13.1 server the player will always see the item as undamaged.

7kasper:
Yes! Work is in progress. This is related to legacy data remappers.

Brokkonaut:
Thx, this works now

PR #910, opened 2018-10-07
[PATCH] Fix complex itemstack remapping.

7kasper:
Kind of band-aid, but this fixes issues such as #900 #905 with minimal changes.

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/itemstack/PreFlatteningItemIdData.java
+++ b/src/protocolsupport/protocol/typeremapper/itemstack/PreFlatteningItemIdData.java
@@ -15,6 +15,7 @@ import protocolsupport.utils.Utils;
 
 public class PreFlatteningItemIdData {
 
+	protected static final int combinedLegacyStoneId = formLegacyCombinedId(1, 0);
 	protected static final int[] toLegacyId = new int[MinecraftData.ITEM_COUNT];
 	protected static final Int2IntMap fromLegacyId = new Int2IntOpenHashMap();
 	protected static void register(String modernKey, int legacyMainId, int legacyData) {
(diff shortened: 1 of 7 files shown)


BBoyJD10:
Can this be merged really need this for the server!

src/protocolsupport/protocol/typeremapper/itemstack/PreFlatteningItemIdData.java
```diff
@@ -46,7 +45,7 @@ public static int getDataFromLegacyCombinedId(int legacyId) {
 	}
 
 	public static int getModernIdByLegacyIdData(int legacyId, int data) {
-		return fromLegacyId.get(formLegacyCombinedId(legacyId, data));
+		return fromLegacyId.getOrDefault(formLegacyCombinedId(legacyId, data), fromLegacyId.getOrDefault(formLegacyCombinedId(legacyId, 0), combinedLegacyStoneId));
```
Shevchik: Remove this hack.

src/protocolsupport/protocol/typeremapper/itemstack/PreFlatteningItemIdData.java
```diff
@@ -24,9 +25,7 @@ protected static void register(String modernKey, int legacyMainId, int legacyDat
 		fromLegacyId.put(combinedLegacyId, modernId);
 	}
 	static {
-		int combinedLegacyStoneId = formLegacyCombinedId(1, 0);
 		Arrays.fill(toLegacyId, combinedLegacyStoneId);
-		fromLegacyId.defaultReturnValue(1);
```
Shevchik: 1 should be replaced with combinedLegacyStoneId

src/protocolsupport/protocol/typeremapper/itemstack/PreFlatteningItemIdData.java
7kasper: How then should we get the modern id of damaged items etc?

7kasper: Ok.. but how can we know which material the item is then? Or shall we just remove the legacydata and make complex remappers for wool etc?

Shevchik: Hmm, that's true, the complex remapper won't be able to deal with that.
I guess we will just leave it like you suggested for now, and will think about better system later.

PR #913, opened 2018-10-15
Change Bone Block mapping to Cobblestone

phase:
Bricks look completely out of place on old clients. Cobblestone looks a
bit better. Quartz would probably look better, but I imagine that both blocks need
to have the same hardness. (Bone & Cobble have a hardness of 2 while Quartz has
a hardness of 0.8)


MrPowerGamerBR:
> (Bone & Cobble have a hardness of 2 while Quartz has
a hardness of 0.8)

Yes, they must have the same hardness to avoid issues when breaking blocks.

krusic22:
Why not go with mossy cobblestone?
It's a lot easier to differentiate others players trash from fossils that way.

PR #937, opened 2018-11-02
v1.7 < 1.13 updates

colinrgodsey:
* biome ids
* item ids
* NPE fixes


src/protocolsupport/protocol/typeremapper/itemstack/ItemStackRemapper.java
```diff
@@ -31,16 +33,18 @@ public static NetworkItemStack remapToClient(ProtocolVersion version, String loc
 
 	public static NetworkItemStack remapFromClient(ProtocolVersion version, String locale, NetworkItemStack itemstack) {
 		if (version == ProtocolVersion.MINECRAFT_PE) {
-			int peCombinedId = PEItems.getModernIdByPEIdData(itemstack.getTypeId(), itemstack.getLegacyData());
-			itemstack.setTypeId(PreFlatteningItemIdData.getIdFromLegacyCombinedId(peCombinedId));
-			itemstack.setLegacyData(PreFlatteningItemIdData.getDataFromLegacyCombinedId(peCombinedId));
+			int modernId = PEItems.getModernIdByPEIdData(itemstack.getTypeId(), itemstack.getLegacyData());
+			itemstack.setTypeId(modernId);
+			itemstack.setLegacyData(0);
```
7kasper: Just remove the set data. Its 0 anyway.

src/protocolsupport/protocol/typeremapper/itemstack/complex/ItemStackComplexRemapperRegistry.java
```diff
@@ -83,6 +83,7 @@ protected static void register(
 			registerToClient(material, enchantfilter, ProtocolVersionsHelper.ALL_PC);
 			registerToClient(material, enchanttolegacyid, ProtocolVersionsHelper.BEFORE_1_13);
 			registerToClient(material, dnametolegacytext, ProtocolVersionsHelper.BEFORE_1_13);
```
7kasper: register with BEFORE_1_13_AND_PE instead of second register (It doesn't make a difference in performance, just is preference)

src/protocolsupport/protocol/typeremapper/itemstack/ItemStackRemapper.java
```diff
@@ -31,16 +30,17 @@ public static NetworkItemStack remapToClient(ProtocolVersion version, String loc
...
+			int modernId = PEItems.getModernIdByPEIdData(itemstack.getTypeId(), itemstack.getLegacyData());
+			itemstack.setTypeId(modernId);
 		} else if (version.isBefore(ProtocolVersion.MINECRAFT_1_13)) {
 			int legacyCombinedId = PreFlatteningItemIdData.formLegacyCombinedId(itemstack.getTypeId(), itemstack.getLegacyData());
 			itemstack.setTypeId(PreFlatteningItemIdData.getModernIdByLegacyCombinedId(legacyCombinedId));
 		} else {
 			itemstack.setTypeId(FlatteningItemId.REGISTRY_FROM_CLIENT.getTable(version).getRemap(itemstack.getTypeId()));
 		}
-		return ItemStackComplexRemapperRegistry.remapFromClient(version, locale, itemstack);
+		itemstack = ItemStackComplexRemapperRegistry.remapFromClient(version, locale, itemstack);
```
7kasper: why not just return it?

src/protocolsupport/protocol/typeremapper/itemstack/ItemStackRemapper.java
```diff
@@ -11,12 +11,11 @@ public static NetworkItemStack remapToClient(ProtocolVersion version, String loc
 		setComplexlyRemapped(itemstack, false);
 		itemstack = ItemStackComplexRemapperRegistry.remapToClient(version, locale, itemstack);
 		itemstack.setTypeId(LegacyItemType.REGISTRY.getTable(version).getRemap(itemstack.getTypeId()));
+
 		if (version == ProtocolVersion.MINECRAFT_PE) {
 			int peCombinedId = PEItems.getPECombinedIdByModernId(itemstack.getTypeId());
 			itemstack.setTypeId(PEItems.getIdFromPECombinedId(peCombinedId));
-			if (!isComplexlyRemapped(itemstack)) {
```
7kasper: I found something wrong with this too yeah. But we cannot just remove it, since it will mess up all our itemstack nbt remappers!

src/protocolsupport/protocol/typeremapper/itemstack/ItemStackRemapper.java
MrPowerGamerBR: My guess? He made it like that so he could added a debug log there, and then he removed the log without inlining the return 😛

7kasper:
See the way I made the biome remapping table.  
Because we have items all with different ids we thought it better to make a big json table. This is a mapping that will be for all future PE versions, because we do Material -> Material remapping on a version basis inside https://github.com/ProtocolSupport/ProtocolSupport/blob/master/src/protocolsupport/protocol/typeremapper/itemstack/LegacyItemType.java Biomes will be better managed with a small version-based remapping table that just defaults to the pc id. (most are the same anyway)

src/protocolsupport/protocol/typeremapper/itemstack/ItemStackRemapper.java
colinrgodsey: ahh @MrPowerGamerBR nailed it lol

src/protocolsupport/protocol/typeremapper/itemstack/ItemStackRemapper.java
colinrgodsey: hmm so i was curious about that. I was setting the typeID to the modern id and the data to 0 before- but the data defaults to 0, so it would have been always 'complex remapped'. but now i removed the setting data to 0 so i dont know :(

colinrgodsey: should i just always run:

```
itemstack.setLegacyData(PEItems.getDataFromPECombinedId(peCombinedId));
```

here? its basically a new item stack, so i think its safe to assume it always needs to be run through the complex remapper.

colinrgodsey: hmm so given that the PC -> PE conversion doesnt do any complex remapping, should be safe to always do this here?
```
setComplexlyRemapped(itemstack, false);
```

colinrgodsey:
should i just do the 'exception' based mapping in java then? only a few biomes changed there

src/protocolsupport/protocol/typeremapper/itemstack/ItemStackRemapper.java
colinrgodsey: hmm things get weird unless i set the data to 0 =\

src/protocolsupport/protocol/typeremapper/pe/PEDataValues.java
```diff
@@ -234,7 +233,18 @@ protected HashMapBasedIdRemappingTable createTable() {
 	public static final IdRemappingRegistry<ArrayBasedIdRemappingTable> BIOME = new IdRemappingRegistry<ArrayBasedIdRemappingTable>() {
 		{
 			registerRemapEntry(Biome.FROZEN_OCEAN, 46, ProtocolVersion.MINECRAFT_PE);
-			registerRemapEntry(Biome.SMALL_END_ISLANDS, Biome.THE_END, ProtocolVersion.MINECRAFT_PE);
+			registerRemapEntry(Biome.SMALL_END_ISLANDS, Biome.OCEAN, ProtocolVersion.MINECRAFT_PE);
```
7kasper: Why ocean as opposed to end?

colinrgodsey: forgot the end exists in pe

PR #941, opened 2018-11-04
Fixes writeable books showing no content

### LINKED ISSUE #931: Unsigned writable books are losing their NBT data when dropped or moved through a container.
Paper dev 409 (Spigot 1.13.2)
ProtocolSupport dev 302

With ProtocolSupport enabled, unsigned writable books lose their NBT data when dropping the book or moving it through containers (hopper -> chest). Sometimes it doesn't even save the book data when clicking the done button (not signing) - it just returns blank when re-opening again. 

Removing ProtocolSupport, unsigned written books preserve data perfectly fine again.

Edit: It might just be a visual issue? Removing ProtocolSupport from my test server after testing with an unsigned book with the text "test" being lost when dropped and picked up again, the book got its text back.

Brokkonaut:
Fixes #931


7kasper:
👍

PR #944, opened 2018-11-09
If PE user tries to connect with a too old client, give correct error message

magicus:
I tried connecting with PE client 1.6, and got the message "Your sever is outdated. Please use 1.7".  The proper message should have been that my *client* is too old, and should be upgraded to 1.7.

I thought I've fixed this before, but that code seems to have gotten rewritten and lost. I know this case is a bit odd in the PS case, since normally "your client is too old" is not even possible to happen with PS.


magicus:
I thought this was a no-brainer. Is there some issue with this fix preventing it from being pushed?

7kasper:
Yes. My chaoticness.

PR #945, opened 2018-11-10
Fix peid for colored stained glass.

magicus:
This solves the mapping issue for stained glass block items.


7kasper:
Thanks magicus!

PR #955, opened 2018-11-24
Reformat blockdefinition with newlines for easier reading.

magicus:
The hugely long line is fine for machine parsing, but hopeless when comparing and trying to find errors with the item json. :-(


7kasper:
Fair enough :)

PR #956, opened 2018-11-24
yivesmirror.com is no more. Replace with working mirror.

magicus:


7kasper:
I think this goes for master too...

PR #960, opened 2018-11-25
Trivial FindBug fix that got lost in the splitup between master and mcpenew

magicus:


7kasper:
Weird. my findbugs didn't report this...

magicus:
That's odd. I ran the FindBugs plugin for IntelliJ. Not sure what version of FindBug it is. But it definitely was dead code.

PR #962, opened 2018-11-25
Entity json data cleanup and bounding box fixes

magicus:
Fix incorrect bounding boxes, and make entitydata JSON easier to update in the future.

* Break out the InventoryFilter from entitydata.json and PEEntityData into entitydata_inventory.json and PEEntityInventoryData. (The filter part does not validate as correct json, which prohibited use by json tools like jq).

* Sort and normalize the json in entitydata.json and entitydata_inventory.json.

* Update entitydata.json with correct values for turtle, parrot and wither skeleton.

Note that the InventoryFilter part seems to not work at the moment, so future cleanup/fixes to this will certainly be needed. For now, I just kept the code as similar as possible while splitting into a separate class.

### DIFF
--- a/resources/resources/pe/entitydata.json
+++ b/resources/resources/pe/entitydata.json
@@ -25,63 +25,6 @@
 			"x": 0.0,
 			"y": 1.1,
 			"z": -0.2
-		},
-		"InventoryFilter": {
-			"Filter": "{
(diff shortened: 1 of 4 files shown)


7kasper:
Ok then :)

PR #966, opened 2018-12-01
Fix a lot of PE item IDs

magicus:
This patch will fix most (hopefully, all) PE item IDs that have a trivial match between PC and PE. Still unhandled is the question on how to deal with items that are only represented on either platform.

Many Bothans died to bring us this information.


magicus:
With the latest patch, any creative items that are only available on PE will be filtered out from the PE GUI. So this solves 50% of the problem of non-matching items. Still not clear what to do with all items that are present on PC only...

7kasper:
@magicus all items only in pc, should be remapped to something similar that does exist in pe, afterwards it is this item that goes through the itemmapping.json. This remapping should be done inside https://github.com/ProtocolSupport/ProtocolSupport/blob/mcpenew/src/protocolsupport/protocol/typeremapper/itemstack/LegacyItemType.java where as you can see all items missing in older minecraft versions are also remapped to something else.

7kasper:
Thanks a lot for this PR!

PR #973, opened 2018-12-06
Even more block remap fixes

magicus:


Artuto:
Apparently Glazed Terracotta is also badly remapped when used in some specific positions

magicus:
@Artuto What colors of glazed terracotta? What positions?

PR #974, opened 2018-12-07
Fix even more block remaps, and a crash while placing bed

magicus:
* All wooden pressure plates, all wooden trapdoors, all prismarine variant slabs.
* Stone and oak buttons got directions fixed.
* Fix so placing a bed does not crash Spigot (v_8_9r1_9r2_10_11_12r1_12r2_13.BlockTileUpdate did not expect a PE version).


src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/BlockChangeMulti.java
```diff
@@ -18,6 +22,14 @@ public BlockChangeMulti(ConnectionImpl connection) {
 
 	protected final TileEntityRemapper tileremapper = TileEntityRemapper.getRemapper(connection.getVersion());
 
+	private static ClientBoundPacketData createPacket(ProtocolVersion version, TileEntity tile) {
+		ClientBoundPacketData serializer = ClientBoundPacketData.create(ClientBoundPacket.PLAY_UPDATE_TILE_ID);
```
7kasper: Huh? This seems like the wrong place for this packet? The problem is that the import used to be wrong I think.
The solution is to fix the import, not make a copy of that method here I think.

magicus:
Yeah, that could very well be the wrong solution. You can remove it from the PR if you wish and push the rest. 

I just redid what the original code did, without the failing version check.

7kasper:
OH nvm i will fix it.

PR #978, opened 2018-12-08
Fix whitespace issues

magicus:
Fix whitespace issues, such as: space after if/switch, empty line consistently before methods, no trailing whitespaces, and some other whitespace-related fixes.

This is part of some kind of effort on getting mcpenew in a slightly better shape for integration.

This patch, while it touches many files, changes nothing but whitespace. It was created by selecting all PE-only files I could locate, asking IntelliJ to fix up formatting, and then manually checking the result of each modified file, and reverting those changes that did not seem reasonable. (And in a few cases, fix the issue that IntelliJ found, but in a different way, e.g. when aligning large blocks of assignments.)


7kasper:
Thanks :)

PR #980, opened 2018-12-09
Fix whitespace issues in master

magicus:
Fix whitespace issues in master (to facilitate further cleanups in mcpenew).

When fixing whitespace issues in mcpenew, I noticed that not all code in master were correctly formatted. To be be able to check that mcpenew does not bring in badly formatted code, it is most helpful to know that master is correctly formatted.

Just like the similar patch for mcpenew I recently submitted, this has been created by using IntelliJ code formatting, and then manually checking all changes, and reverting those that did not make sense.

The majority of the changes is inserting an empty line in front of methods. Another common change is to remove double spaces, or replace leading spaces with tabs. No other changes than whitespace has been made in this patch.


7kasper:
I think there's some extra lines that we don't really need or want, but in general there's a lot more fixes than  unnecesarry optimisation. So if @Shevchik doesn't mind I think we can merge this.

7kasper:
image
Bescides those points we still might want to adress and clean this up some time. For now I will close the PR.

PR #982, opened 2018-12-09
Refactor writeUUID() to take a version, in preparation for mcpenew integration

magicus:
I'm looking at the diff between master and mcpenew, and trying to minimize it, from both sides. 

This is a simple refactoring that adds a version argument to writeUUID().


Shevchik:
A better solution would be a separate method called wirteUUIDLe.
ProtocolSupport slowly moves away from having a version based write in serializers. Instead the middle packets should know how exactly we need to write complex type to stream.

PR #983, opened 2018-12-09
Refactor the MiddleBlocks to have a create() method, which is called by the constructor.

magicus:
I'm looking at the diff between master and mcpenew, and trying to minimize it, from both sides.

This is a simple refactoring that breaks out the actual work of the constructor into a create() method, that can be independently called, for all middle blocks, which is needed by PSPE.


7kasper:
👍

PR #984, opened 2018-12-09
Make creative inventory look a bit less messed up.

magicus:
This is not the complete, correct solution. As stated in the comment, that means parsing the base-64 formatted NBT tag. However, this stops the creative inventory from becoming spammed with enchanted books without enchantments, etc.


7kasper:
So yeah the nbt tags are just taken from PocketMine NP creativeitems json. We might want a different custom implementation by ourselfs in the future, but for now this was the easiest way to get stuff working quickly. I will accept this, since that whole system is a bit of an experiment anyways.

colinrgodsey:
My new branch *should* have the fix for enchanted books. I'm not 100% sure i did it right, but i think i did (I really should have looked at the NBT in the creative b64 payload). I'm just using the normal 'ench' field in NBT for the books. They look right and work, but it still seems suspicious lol. I did notice even in single player, all enchant books show as 'item' category, and not the real category for the what they should be bound to... so not sure

PR #985, opened 2018-12-10
Create PE-specific version of writeUUID

magicus:
Create PE-specific version of writeUUID, to remove PE-specific changes to non-PE code.

This was prompted by Shev's response in https://github.com/ProtocolSupport/ProtocolSupport/pull/982.

I'd like some feedback on the fix in "MiddleSpecate.java" (also! Someone should fix the typo of that class name!). I assume I can remove the ProtocolType check and just assume PC, but I'd like to hear someone confirm it. (I'm not sure really why the middle classes seem to be doing middleimpl stuff...)

With this patch, the diff between master and mcpenew outside PE files will be reduced (but not zero).

### DIFF
--- a/src/protocolsupport/protocol/packet/middle/serverbound/play/MiddleSpecate.java
+++ b/src/protocolsupport/protocol/packet/middle/serverbound/play/MiddleSpecate.java
@@ -2,6 +2,7 @@ package protocolsupport.protocol.packet.middle.serverbound.play;
 
 import java.util.UUID;
 
+import protocolsupport.api.ProtocolType;
 import protocolsupport.protocol.ConnectionImpl;
 import protocolsupport.protocol.packet.ServerBoundPacket;
 import protocolsupport.protocol.packet.middle.ServerBoundMiddlePacket;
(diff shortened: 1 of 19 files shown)


src/protocolsupport/protocol/packet/middle/serverbound/play/MiddleSpecate.java
```diff
@@ -21,7 +22,12 @@ public MiddleSpecate(ConnectionImpl connection) {
 	@Override
 	public RecyclableCollection<ServerBoundPacketData> toNative() {
 		ServerBoundPacketData creator = ServerBoundPacketData.create(ServerBoundPacket.PLAY_SPECTATE);
-		MiscSerializer.writeUUID(creator, connection.getVersion(), entityUUID);
+		// FIXME: Maybe we can assume PC here..? Only implemented in v_8_9r1_9r2_10_11_12r1_12r2_13
```
7kasper: better yet. This is serverbound, the function therefore is always targetted at the latest pc version. Writing PE uuid in here would just crash the connection.

magicus:
Thanks! Will update MiddleBlock next time I have some PSPE dev time.

And thanks to you for opening the PR of mcpenew to master. It triggered me to look things over.

magicus:
Finally fixed. :)

7kasper:
Awesome!

PR #992, opened 2018-12-17
Updates due to Spigot's NMS changes

MrPowerGamerBR:
Untested because my computer is freezing like crazy

Updated Paper version, I also needed to rename `tick()` to `loginTick()` because `ITickable`'s method was renamed to... `tick()`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ librariesShadeDirectory.mkdirs()
 task updateRuntimeLibraries(type: UpdateLibrariesTask) {
 	directory = librariesRuntimeDirectory
 	libraries = [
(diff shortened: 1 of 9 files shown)


MrPowerGamerBR:
@Shevchik what name should I change to? Minecraft version wasn't bumped, and the previous JAR was `spigot-1.13.2.jar`, shouldn't it be `spigot-1.13.2-1.jar`? (after all, the latest Minecraft version is still 1.13.2, it was just Spigot that renamed some fields and broke ProtocolSupport)

Shevchik:
It should spigot-1.13.2-2.jar. The problem is that we had a name of spigot-1.13.2-1.jar before, then it was changed to spigot-1.13.2.jar (which was invalid).

PR #1020, opened 2019-02-07
[MCPE] 1.9 + 1.10 + upstream + new raknet

colinrgodsey:
also removes 1.7


src/protocolsupport/protocol/serializer/ItemStackSerializer.java
```diff
@@ -243,8 +243,15 @@ public static NBTCompound readTag(ByteBuf from, boolean varint, ProtocolVersion
 			} else if (isUsingPENBT(version)) {
 				if (!varint) { // VarInts NBTs doesn't have length
 					final short length = from.readShortLE();
-					if (length <= 0) {
+					if (length == 0) {
 						return null;
+					} else if (length == -1) {
+						final int numNBT = from.readByte(); //TODO: wtf, why multiple NBT?
```
Shevchik: Should throw exception on multiple nbt tags instead

PR #1045, opened 2019-03-21
Fix remaps for TNT, glazed terracotta, levers

magicus:


colinrgodsey:
same here, this looks good to me. lets get it merged in

PR #1036, opened 2019-02-28
Refactor name tag remapping

MrPowerGamerBR:
Now nameplates behave like Minecraft PC.

1. For mobs: You can only see nameplates of a entity if you are looking at it.
2. For players: Nameplates will be displayed correctly

Note: Toggling nameplates with Citizens is still a bit wonky (only goes away when relogging/reloading the entity)

PR #1052, opened 2019-04-02
[MCPE] player nametag fix

colinrgodsey:
(cherry picked from commit eaddb0c2917dbdf86ca0510b205e77506a8a0d43)

Player nametag fix, thanks to @MrPowerGamerBR !

Replaces #1036

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/entity/metadata/types/special/PlayerEntityMetadataRemapper.java
+++ b/src/protocolsupport/protocol/typeremapper/entity/metadata/types/special/PlayerEntityMetadataRemapper.java
@@ -1,23 +1,53 @@
 package protocolsupport.protocol.typeremapper.entity.metadata.types.special;
 
 import protocolsupport.api.ProtocolVersion;
+import protocolsupport.api.chat.components.BaseComponent;
 import protocolsupport.protocol.packet.middleimpl.clientbound.play.v_pe.EntityMetadata.PeMetaBase;
 import protocolsupport.protocol.typeremapper.entity.metadata.DataWatcherObjectRemapper;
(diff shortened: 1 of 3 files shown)


src/protocolsupport/protocol/typeremapper/entity/metadata/types/special/PlayerEntityMetadataRemapper.java
```diff
@@ -1,23 +1,53 @@
...
-		addRemap(new PeSimpleFlagAdder(new int[] {PeMetaBase.FLAG_ALWAYS_SHOW_NAMETAG}, new boolean[] {true}), ProtocolVersionsHelper.ALL_PE);
+		addRemap(new DataWatcherObjectRemapper() {
+			@Override
+			public void remap(NetworkEntity entity, ArrayMap<DataWatcherObject<?>> original, ArrayMap<DataWatcherObject<?>> remapped) {
+				// = PE Nametag =
+				// Players has some small edge cases with name tags
+				Optional<DataWatcherObjectOptionalChat> nameTagWatcher = DataWatcherObjectIndex.Entity.NAMETAG.getValue(original);
+				Optional<DataWatcherObjectBoolean> nameTagVisibilityWatcher = DataWatcherObjectIndex.Entity.NAMETAG_VISIBLE.getValue(original);
+				boolean hasCustomNameTagMetaValue = nameTagWatcher.isPresent();
+				boolean hasNameTagPersistentMetaValue = nameTagVisibilityWatcher.isPresent();
```
Shevchik: That should be always true for players. We should ignore hacks with player visibility which work onlt on some versions actually.

Artuto:
This apparently does not work with plugins like NametagEdit for players (which on PE are just white). Dont know if they do any black magic so it breaks, PE does not support colors there or there's a bug in PSPE.

PR #1053, opened 2019-04-03
[MCPE] Best possible dimension handling possible.

colinrgodsey:
Just... let it happen, its fine.


Shevchik:
As i already said - no. I can accept relying on bukkit to send chunk publisher, nut not the dimension switch.

Remu1012:
why not it if works it works :D

Shevchik:
Because just works is not good enough.

colinrgodsey:
The ordering doesnt make sense if you trigger change dimension from the packet. I can add extra state handling to try and trigger some of these actions when it gets the change dimension packet, but its just going to make the code more brittle and stupid. this is the most correct implementation. its not that it "just works", its that it works correctly, and in a simple, straight-forward, and less error-prone way.

colinrgodsey:
dimension switch driven by the packet now, uses the cached dimension set by the PEChunkPublisher to determine if it has to send fake chunks and lock
still works just as well 👍 :

src/protocolsupport/protocol/storage/netcache/MovementCache.java
```diff
@@ -1,15 +1,27 @@
...
 public class MovementCache {
 
 	protected static final double acceptableError = 0.1;
+	protected static final int peIncreasedLeniencyMillis = 1000;
 
 	protected double x;
 	protected double y;
 	protected double z;
 	protected int teleportConfirmId;
+	protected double peClientX;
```
Shevchik: Move it all back down

src/protocolsupport/protocol/pipeline/version/v_pe/PEPacketDecoder.java
```diff
@@ -92,6 +92,14 @@ public int readPacketId(ByteBuf from) {
 		return sReadPacketId(from);
 	}
 
+	public static int sPeakPacketId(ByteBuf from) {
```
Shevchik: peek

Shevchik:
Now it needs the implementation of single send of LoginSuccess PLAYER_SPAWN, so we can actually lock queue using this packet, not the separate custom payload message.

Shevchik:
The custom payload change still needs to be removed. Or updated to use pre13 converter logic for both client -> server and server -> client.

Shevchik:
Soo, let's try this.

PR #1056, opened 2019-04-04
Make trading great again.

magicus:
(The critical part with the extra byte is thanks to Yesdog, the refactoring is my fault.)

### DIFF
--- a/src/protocolsupport/protocol/serializer/ItemStackSerializer.java
+++ b/src/protocolsupport/protocol/serializer/ItemStackSerializer.java
@@ -287,18 +287,6 @@ public class ItemStackSerializer {
 		return ItemStackRemapper.remapToClient(version, locale, itemstack);
 	}
 
-	public static NBTCompound itemStackToPENBT(ProtocolVersion version, String locale, NetworkItemStack itemstack) {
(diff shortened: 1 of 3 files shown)


src/protocolsupport/protocol/serializer/ItemStackSerializer.java
```diff
@@ -285,6 +287,18 @@ public static NetworkItemStack remapItemToClient(ProtocolVersion version, String
 		return ItemStackRemapper.remapToClient(version, locale, itemstack);
 	}
 
+	public static NBTCompound itemStackToPENBT(ProtocolVersion version, String locale, NetworkItemStack itemstack) {
```
Shevchik: The proper place for it is the CommonNBT class

magicus: @Shevchik Indeed! The function was very similar to the pre-existing serializeItemStackToNBT. I renamed it to serializeItemStackToPENBT to highlight this similarity.

It is possible that the two functions could be united more, but they seem to do slightly different remapping and stuff, so I thought at the point trying to unite what was similar would just make it harder to read.

PR #1064, opened 2019-04-12
The Update Soundatic -- fix PE sound remaps.

magicus:
Apart from fixing PE sound mapping, I made a few non-PE-related fixes: Removed mappings for which the server sound was non-existant, and fixed typos. (There is some probable typos left for non-PE mappings, but they are on the client side so I can't verify that they are not indeed intentional, if the typo is in the PC client itself, like "mob.cat.hitt").

I have verified that the following properties hold for the mappings:
 * All PC server sounds that I could reasonably figure out what to map to has a mapping
 * All PE client sounds that should reasonably be mapped to has a mapping

Note that apart from a few samples, I have not actually *verified* the mappings, I have just used common sense in connecting what seemed to imply the same sound.

Also note that PE client is doing funky stuff with sounds. I think it's playing a few player-initiated sounds by it's own account. On the other hand, some player-initiated sounds (like breaking blocks) is not heard, regardless of the correct mapping, if the breaking is done by the player herself, and not by another player. Needs some further examination.


Shevchik:
Fixing or changing pc related remaps are not allowed in pe branch

magicus:
Do you want me to revert those fixes, or separate them out into a PR targeting master? I could do either, you decide.

Shevchik:
Revert, the main branch is under code freeze until upda to 1.14

magicus:
Done.

Shevchik:
That warning still needs to be removed. Also stop using ProtocolVersion::isPe

magicus:
I thought I could keep the warning as long as I made it PE-specific? I'd like to avoid getting duplicates back in the PE mapping. But sure, I can move it over to my debug-branch.

What should I use instead of isPe?

PR #1065, opened 2019-04-12
Fix problem with updating tile entity cache for PE.

magicus:


src/protocolsupport/protocol/typeremapper/chunk/ChunkTransformerPE.java
```diff
@@ -62,8 +70,8 @@ public void writeLegacyData(ByteBuf chunkdata) {
 		}
 	}
 
-	protected int getBlockState(int section, BlockStorageReader storage, int x, int y, int z) {
-		return getBlockState(section, storage, (y << 8) | (z << 4) | (x));
+	private int getPcIndex(int x, int y, int z) {
```
Shevchik: We use protected everywhere we can to avoid generation of accessor methods. Also this method can now be static.

src/protocolsupport/protocol/typeremapper/chunk/ChunkTransformerPE.java
```diff
@@ -32,11 +32,19 @@ public void writeLegacyData(ByteBuf chunkdata) {
...
+				int peIndex = 0;
 				for (int x = 0; x < 16; x++) { for (int z = 0; z < 16; z++) { for (int y = 0; y < 16; y++) {
-					if (PEBlocks.isPCBlockWaterlogged(getBlockState(i, storage, x, y, z))) { waterstorage.setBlockState(blockIndex, 1); }
-					blockstorage.setBlockState(blockIndex++, palette.getRuntimeId(PEBlocks.getPocketRuntimeId(blockDataRemappingTable.getRemap(getBlockState(i, storage, x, y, z)))));
+					int pcIndex = getPcIndex(x, y, z);
+					int blockstate = storage.getBlockData(pcIndex);
+
+					if (PEBlocks.isPCBlockWaterlogged(blockstate)) { waterstorage.setBlockState(peIndex, 1); }
+					// Update tile entity cache
+					processBlockData(i, pcIndex, blockstate);
```
Shevchik: Should be moved up a line, so it is positioned just after blockstate get.

PR #1071, opened 2019-04-13
Fix so banner items are not always white and bed items are not always red.

magicus:
I found and fixed the issue! Also fixed red beds, which were plagued by the same problem.

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/itemstack/LegacyItemType.java
+++ b/src/protocolsupport/protocol/typeremapper/itemstack/LegacyItemType.java
@@ -6,6 +6,7 @@ import java.util.List;
 import org.bukkit.Material;
 
 import protocolsupport.api.MaterialAPI;
+import protocolsupport.api.ProtocolType;
 import protocolsupport.api.ProtocolVersion;
 import protocolsupport.protocol.typeremapper.block.LegacyBlockData;
 import protocolsupport.protocol.typeremapper.utils.RemappingRegistry.IdRemappingRegistry;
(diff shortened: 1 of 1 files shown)


src/protocolsupport/protocol/typeremapper/itemstack/LegacyItemType.java
```diff
@@ -24,6 +25,16 @@
 		{
 			applyDefaultRemaps();
 		}
+
+		protected boolean noDefaultRemap(Material materialTo, ProtocolVersion version) {
```
Shevchik: Just remap them again instead of prohibiting it.

magicus: @Shevchik I started out doing that, but there's a lot of remaps; all beds, and all banners (and I have a vague feeling there is one more item with the same issue that I've forgotten about right now) -- I'd need to map "BLACK_BED" to "BLACK_BED", etc, for all of the remaps. 

Apart from lots of code, it looked really odd -- why remap stuff to themselves? The default should be, as elsewhere, that if a thing is the same in both the server and the client, then no remap should be necessary.

At first I thought a better way was to register the block remaps with an extra argument (noItemRemap = true) or something like that, but I couldn't really find a fitting way to propagate this info.

I could, of course, thinking loud right now, use that to populate a list that is read in LegacyItemType. Do you think that would be better?

magicus: ... or be more explicit and let the register functions in LegacyBlockData register both to a block remap and an item remap, to two separate remaptables (and skip the item remap if told so by an argument).

Shevchik: Your current setup just blocks registry reset.
There is already a todo here for actually adapting the block remap table.

PR #1073, opened 2019-04-18
Add support for tab completion on PE

magicus:
This is done by reading and parsing the DeclareCommands packet from the PC server, and converting into the corresponding AVAILABLE_COMMANDS packet for PE. Most, but not all, functionality is kept.

### DIFF
--- a/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
+++ b/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
@@ -284,7 +284,7 @@ public class DeclareCommands extends MiddleDeclareCommands {
 		});
 		argTypeReaders.put("minecraft:score_holder", from -> {
 			// The "multiple" boolean is true if multiple, false if single.
-			byte multiple = from.readByte();
+			boolean multiple = from.readBoolean();
 		});
 	}
 
@@ -443,7 +443,7 @@ public class DeclareCommands extends MiddleDeclareCommands {
 
 		// The "flag" design is really... odd. Bedrock Edition engineers. Don't ask.
 		serializer.writeIntLE(flag);
-		serializer.writeByte(0); // Boolean IS_OPTIONAL (1 = true). For now, call everything compulsory.
+		serializer.writeBoolean(false); // Boolean IS_OPTIONAL. For now, call everything compulsory.
 		serializer.writeByte(0); // Flags? Always 0.
 	}
 
--- a/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
+++ b/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
@@ -166,13 +166,7 @@ public class DeclareCommands extends MiddleDeclareCommands {
 		 * Register a literal in the internal literal registry if it was new, and return its index value.
 		 */
 		public int registerLiteral(String literal) {
-			Integer nextValue = literalRegistry.size();
-			Integer existingValue = literalRegistry.putIfAbsent(literal, nextValue);
-			if (existingValue != null) {
-				return existingValue;
-			} else {
-				return nextValue;
-			}
+			return literalRegistry.computeIfAbsent(literal, k -> literalRegistry.size() );
 		}
 
 		/**
--- a/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
+++ b/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
@@ -14,8 +14,10 @@ import protocolsupport.utils.recyclable.RecyclableSingletonList;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.function.Consumer;
 
 public class DeclareCommands extends MiddleDeclareCommands {
@@ -147,8 +149,7 @@ public class DeclareCommands extends MiddleDeclareCommands {
 
 	private static class PECommandsStructure {
 		private Map<CommandNode, List<List<PEArgumentNode>>> overloadsRegistry = new HashMap<>();
-		private Map<String, Integer> literalRegistry = new HashMap<>();
-		private String[] literalArray;
+		private LinkedHashMap<String, Integer> literalRegistry = new LinkedHashMap<>();
 
 		/**
 		 * Register a list of overloads for a top-level node.
@@ -162,35 +163,23 @@ public class DeclareCommands extends MiddleDeclareCommands {
 		}
 
 		/**
-		 * Register a literal in the internal literal array if it was new, and return it's index value.
+		 * Register a literal in the internal literal registry if it was new, and return its index value.
 		 */
 		public int registerLiteral(String literal) {
-			if (literalArray != null) {
-				throw new IllegalStateException("Cannot register literals after getLiteralArray() is called");
+			Integer nextValue = literalRegistry.size();
+			Integer existingValue = literalRegistry.putIfAbsent(literal, nextValue);
+			if (existingValue != null) {
+				return existingValue;
+			} else {
+				return nextValue;
 			}
-
-			Integer index = literalRegistry.get(literal);
-			if (index == null) {
-				index = literalRegistry.size();
-				literalRegistry.put(literal, index);
-			}
-			return index;
 		}
 
 		/**
-		 * Return an array of all registered literals. All strings in the array are unique.
+		 * Return an ordered set of all registered literals.
 		 */
-		public String[] getLiteralArray() {
-			// Not thread safe...
-			if (literalArray == null) {
-				// Convert literalRegistry to proper array using assigned indices
-				literalArray = new String[literalRegistry.size()];
-				for (Map.Entry<String, Integer> entry : literalRegistry.entrySet()) {
-					literalArray[entry.getValue()] = entry.getKey();
-				}
-			}
-
-			return literalArray;
+		public Set<String> getLiteralArray() {
+			return literalRegistry.keySet();
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


Shevchik:
private -> protected as we do elsewhere.
Also replace elseif chains with maps.

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
```diff
@@ -0,0 +1,475 @@
...
+
+		/**
+		 * Register a literal in the internal literal array if it was new, and return it's index value.
+		 */
+		public int registerLiteral(String literal) {
+			if (literalArray != null) {
+				throw new IllegalStateException("Cannot register literals after getLiteralArray() is called");
+			}
+
+			Integer index = literalRegistry.get(literal);
```
Shevchik: computeIfAbsent?

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
```diff
@@ -0,0 +1,475 @@
...
+		}
+
+		public int getNameIndex() {
+			return nameIndex;
+		}
+	}
+
+	private static class PECommandsStructure {
+		private Map<CommandNode, List<List<PEArgumentNode>>> overloadsRegistry = new HashMap<>();
+		private Map<String, Integer> literalRegistry = new HashMap<>();
```
Shevchik: This can be just a Set based on LinkedHashMap. That way you don't have to maintain an index. Will also allow just passing Collection to writeLiteral methods.

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/StartGame.java
```diff
@@ -133,6 +133,10 @@ public StartGame(ConnectionImpl connection) {
 		VarNumberSerializer.writeSVarInt(chunkradius, (int) Math.ceil((Bukkit.getViewDistance() + 1) * Math.sqrt(2)));
 		packets.add(chunkradius);
 
+		ClientBoundPacketData enableCommandsPacket = ClientBoundPacketData.create(PEPacketIDs.SET_COMMANDS_ENABLED);
+		enableCommandsPacket.writeByte(1); // boolean true = enable commands
```
Shevchik: writeBoolean?

magicus: I thought we didn't have one... Thought I looked for it, didn't find it, looked how others did it, and settled for writeByte. I see now you're correct; will fix. I should probably also look at more places were we use writeByte currently (in a separate fix).

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
magicus: I'm not sure I follow you 100%. My main constraint here is that I need to assign each and every unique literal a value; these values must be consecutive, and I must at a later point be able to do two-way conversions (i.e. string->num and num->string). I also prefer to do these conversations as efficient as possible.

I can't really see how a Set can do that.

However, for the overloads, they could be Set<List<PEArgumentNode>> since the order between overloads do not (or should not) matter, only between arguments in each overload. However, I prefer to keep them in order; I have not tested but presumably the PE client is brain dead and presents them in the order I send them.

Shevchik: Based on the code i seem, you don't need string->num and num->string convert.
With linkedhashmap the iteration order index will be your index (which can be used in writeLiteral*) and you can just add the new literal to the map if it is not already present, so the order will stay.

magicus: Yes, it might be possible to use a LinkedHashMap instead. I'd forgotten about that data type. However, I'm not sure how much it improves the code. We'll still need to use the Map part of it to map String to Integer, mapping in an increasing value as is done now, and then I can iterate over the List part in insertion order -- this will happen to correspond to the mapped Integers, but this will not be due to the design of the datatype, but due to careful usage.

Ah heck. I should probably subclass LinkedHashMap and make this functionality explicit.

Shevchik: Yeah, the usage of LinkedHashMap in such manner will require a comment about that. That actually decrease code readability, yet improves performance.

magicus:
@Shevchik Looking better now?

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
```diff
@@ -0,0 +1,479 @@
...
+		}
+
+		public List<List<PEArgumentNode>> getOverloads(CommandNode node) {
+			return overloadsRegistry.get(node);
+		}
+
+		/**
+		 * Register a literal in the internal literal registry if it was new, and return its index value.
+		 */
+		public int registerLiteral(String literal) {
```
Shevchik: computeIfAbsent

magicus: Thanks. Last time I programmed Java seriously was prior to lambdas, so I'm still catching up on all new functionality in the core classes.

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
```diff
@@ -0,0 +1,473 @@
...
+			List<PEArgumentNode> argumentList = new ArrayList<>();
+			for (CommandNode node : nodePath) {
+				PEArgumentNode peNode = new PEArgumentNode(node, peStruct);
+				argumentList.add(peNode);
+			}
+			overloads.add(argumentList);
+		} else {
+			for (int childNodeIndex : currentNode.getChildren()) {
+				CommandNode childNode = allNodes[childNodeIndex];
+				List<CommandNode> newPath = new ArrayList<>(nodePath);
```
Shevchik: Might be worth investigating if clone can be avoided.
For example you may add a childnode and then remove it after calling walkNode.

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
```diff
@@ -0,0 +1,473 @@
...
+			// See FLAG_ENTITY_AMOUNT_IS_SINGLE and FLAG_ENTITY_TYPE_IS_PLAYER.
+			byte flag = from.readByte();
+		});
+		argTypeReaders.put("minecraft:score_holder", from -> {
+			// The "multiple" boolean is true if multiple, false if single.
+			boolean multiple = from.readBoolean();
+		});
+	}
+
+	private static String readArgType(ByteBuf from) {
```
Shevchik: Collapse into single line and move the comment

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
```diff
@@ -0,0 +1,473 @@
...
+			topLevelNodes[i++] = allNodes[index];
+		}
+	}
+
+	private static CommandNode readCommandNode(ByteBuf from) {
+		byte flags = from.readByte();
+		boolean isPathEnd;
+		int redirect;
+		int[] children = ArraySerializer.readVarIntVarIntArray(from).clone();
+		if ((flags & FLAG_HAS_REDIRECT) != 0) {
```
Shevchik: ternary

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
```diff
@@ -0,0 +1,473 @@
...
+			redirect = VarNumberSerializer.readVarInt(from);
+		} else {
+			redirect = -1;
+		}
+		String name;
+		String argType;
+		String suggestion;
+
+		isPathEnd = ((flags & FLAG_IS_COMMAND_PATH_END) != 0);
```
Shevchik: can be actually inlined to constructor call

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
```diff
@@ -0,0 +1,473 @@
...
+		for (int index : topLevelNodeIndices) {
+			topLevelNodes[i++] = allNodes[index];
+		}
+	}
+
+	private static CommandNode readCommandNode(ByteBuf from) {
+		byte flags = from.readByte();
+		boolean isPathEnd;
+		int redirect;
+		int[] children = ArraySerializer.readVarIntVarIntArray(from).clone();
```
Shevchik: Don't think we actually need a clone there

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
magicus: Maybe, and maybe not. I think there's a limit to how much effort I'm willing to spend on micro-optimizations on this sent-once-per-login packet...

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
magicus: But why?

src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_pe/DeclareCommands.java
magicus: How's that helping anything but changing code for the sake of change?

Shevchik: Looks ugly, like if we were writing c. Plus when inspecting code initially it all looks like you have muptiple places where this is assigned, why it actually isn't. Which also requires more time to read and understand.

PR #1074, opened 2019-04-18
Fix colliding servers

magicus:
The unknown field in getServerInfo() in PEProxyServerInfoHandler is indeed some kind of ID. It must be unique per server instance, otherwise PE client will lump all servers on the LAN with the same ID together. Without this fix, it was not possible to run two different instances of PSPE on a LAN (and see them as separate).

### DIFF
--- a/src/protocolsupport/zplatform/impl/pe/PEProxyServerInfoHandler.java
+++ b/src/protocolsupport/zplatform/impl/pe/PEProxyServerInfoHandler.java
@@ -1,6 +1,7 @@
 package protocolsupport.zplatform.impl.pe;
 
 import java.text.MessageFormat;
+import java.util.Random;
 import java.util.concurrent.Executor;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadPoolExecutor;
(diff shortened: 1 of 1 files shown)


src/protocolsupport/zplatform/impl/pe/PEProxyServerInfoHandler.java
```diff
@@ -30,6 +31,7 @@
 
 	private static final int statusThreads = JavaSystemProperty.getValue("pestatusthreads", 2, Integer::parseInt);
 	private static final int statusThreadKeepAlive = JavaSystemProperty.getValue("pestatusthreadskeepalive", 60, Integer::parseInt);
+	private static final int randomId = new Random().nextInt();
```
Shevchik: Use ThreadLocalRandom. And it should be a string from a start to save time on String.valueOf calls

magicus: I hardly believe this matters. We're doing repeated regexp operations on the PE version string... If we were worried about extreme performance in this method, we should start there. And it's very much unclear that creating a thread local random PRNG for a single call at startup really isn't a regression, rather than an optimization.

Shevchik: We don't do regex on pe version tho, or at least not supposed to. So it does matter even if it's a simple thing. Also you don't save anything on not calling threadlocalrandom, because the seed for it is actually stored in the thread.

PR #1081, opened 2019-04-30
Fixed rabbit size in PSPE

KernelFreeze:
Hello! I added a new metadata remap to send to clients the right size of rabbits. If another change is needed, or something can be done in a better way, don't hesitate to tell me.

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/entity/metadata/types/base/BaseEntityMetadataRemapper.java
+++ b/src/protocolsupport/protocol/typeremapper/entity/metadata/types/base/BaseEntityMetadataRemapper.java
@@ -99,13 +99,11 @@ public class BaseEntityMetadataRemapper extends EntityMetadataRemapper {
 			mapPut.accept(PeMetaBase.MAX_AIR, new DataWatcherObjectShortLe(300));
 			// = PE Bounding Box =
 			PEEntityData pocketdata = PEDataValues.getEntityData(entity.getType());
-			if (pocketdata == null) {
+			if (pocketdata == null || pocketdata.getBoundingBox() == null) {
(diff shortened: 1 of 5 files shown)


magicus:
It seems complex to just fix the rabbit baby size... I imagine that it is not really just rabbit babies that are off, but all babies, only that it is more visible with rabbits. So I think the solution should be more general. I've always imagined that we'd need to encode the baby size in entitydata.json.

WesleyVanNeck:
@magicus its all rabbits not only babies mostly adults There like from 1.8 size there like mc java small

WesleyVanNeck:
image

@magicus

magicus:
Hm. It still seems overly messy for just rabbits. 

Any idea why this should be needed for just rabbits, but no other mobs? 

Where does the 0.3 and 0.55 factors come from? 

Why the null check for the bounding box? Either we have a bounding box for rabbits (we should!) or we don't, in case something is messed up with entitydata.json, and we should not just ignore it.

I really appreciate that you try to fix this, I'm just not really comfortable with the fix.

KernelFreeze:
Other baby mobs are already working by default, only rabbits need this fix for some reason. I took the sizes from the wiki, maybe it would be better to get them from a constant or even from entitydata.json, you tell me what you prefer, and I can change it.

Also, I don't think the null check is necessary, but I saw that you used it for Giants, I took it from there.

magicus:
Ok, we'll just have to accept that rabbits are different. In that case, I think the fix is good enough. It still has a bit of a "patchy" feel to it, but then again, so has a lot of the PE code :-).

Shevchik:
Missing bounding box should not be ignored, and it should fail in that case.
Giant should have null check removed too if it exists.

KernelFreeze:
Like this?

PR #1082, opened 2019-04-30
Fixed strange behavior in mobs movement.

KernelFreeze:
With this commit, now entities use relative movement so it is no longer buggy. Related to ProtocolSupport/ProtocolSupportPocketIssues#65.


KernelFreeze:
Maybe something like this?

Shevchik:
The problem with current implementation is that it only tracks changes caused by entity teleport, and not by entity relative move (which ain't implemented right now). While protocolsupport entity tracker only sends teleports for entities, plugins can still send relative movement.
So relative move translation needs to be implemented first, and after that this commit will require changes to account for relative movement.

magicus:
@Shevchik I understand your concern for proper support for plugins, and I agree that it is important.

Nevertheless, this fix still seems like a good partial step in it's own right. For users running PSPE on a server without plugins that use relative movement, this will be a complete fix for the jerky movement (which really is annoying! :-().

I propose that we open an enhancement issue to track the need to implement proper relative move translation, and integrate this bug fix as it is. Sounds reasonable?

Shevchik:
Then TODO need to be added to position cache with a note about also caching on relative move packets.

KernelFreeze:
Ok, I added a note, I think I can take care of implementing it later

PR #1097, opened 2019-05-28
Update to 1.14.2

Brokkonaut:


LinzN:
Hey, is it possible that signs are invisible in your 1.14.2 version?

sachk:
@LinzN There's loads of bugs with all of the 1.14 branches. Don't expect anything close to a bug-free experience. You can use paperspigot 1.13.2 with the 1.13.2 protocolsupport and viaversion which works very well as an alternative, you can even use the PE branch like that. However, 1.14's new additions don't work.

I'm hoping the 1.14 PE  branch will become much more stable soon, but from what I've seen fully fixing PE support is going to require a lot of changes.

PR #1103, opened 2019-06-04
Filter invalid block positions

### LINKED ISSUE #1098: 1.14.1: Client gets kicked when breaking blocks
When breaking several blocks within a short time, the client gets kicked.

For example in a field with much long grass and the player must be in survival mode - when the player is now breaking several of the long grasses within a short time (left mouse button pressed and look around) he gets kicked.

Debug log: https://pastebin.com/raw/sV4WhcsG

Brokkonaut:
Fixes #1098


Shevchik:
The y coord can't be negative (due to encoding) so no actual need for checking if it is < 0

Brokkonaut:
Ok, I changed that

PR #1111, opened 2019-06-16
[MCPE] fixes a kick on unknown sound key

InsaneLegacy:


Shevchik:
That's not how you are supposed to do that.

WesleyVanNeck:
@Shevchik how then? i dont see a other way

WesleyVanNeck:
and it doesnt break anyting ¯\_(ツ)_/¯

PR #1115, opened 2019-06-22
Fix invisible armor stand and possibly others

EOT3000:
Warning, uses nms. 
Can probably make Hologrsms work by using areas effect clouds


Shevchik:
NMS is not allowed. Also PC does have the invisible metadata just in other place.

PR #1118, opened 2019-06-25
Update to 1.14.3

Brokkonaut:


Speentie8081:
How stable is it?

Brokkonaut:
I did a test with 1.12.2, 1.13.2, 1.14.2 and 1.14.3 clients and the changed features in 1.14.3. Don't know if I missed some change, but I think it is somewhat stable.

Speentie8081:
[ProtocolSupportBungee 1.14.3 available on Jenkins](https://ci.skorrloregaming.com/job/ProtocolSupportBungee-1.14.3/)
[ProtocolSupport 1.14.3 available on Jenkins thanks to Brokkonaut](https://ci.skorrloregaming.com/job/ProtocolSupport-1.14.3/)

Xuwznln:
1.14.3 cant open village trade gui but 1.14.2 can. Maybe a bug.

Brokkonaut:
> 1.14.3 cant open village trade gui but 1.14.2 can. Maybe a bug.

I cannot confirm this. I tested with 1.14.3 clients and 1.14.2 clients and could trade with both normal villagers and wandering trader villagers. And I could not find a bug in the code. Maybe you have other plugins blocking the trade or the villager did not have a profession?

Or did you try to open the gui by using a plugin?  If so maybe it is a spigot/paper bug? Did you test without PS?

Xuwznln:
thanks for your reply. I will test it again in detail tomorrow morning.

drtshock:
Running this on Minehut on more than 2000 servers since this morning with no issues reported yet.

Shevchik:
Current villager data serializer is really fucked up. I should probably write some sort of serializer based on version selector instead of this if chain garbage.

Xuwznln:
Today I tested again, no problem. uhm...thats werid.

Speentie8081:
Yeah my players are reporting issues with villagers

sachk:
Same villager problem here

Brokkonaut:
What exactly is wrong (no window, traded do not work, ...)? Is there any exception in the log? Are there ways to reproduce it? Can you reproduce it without any other plugins? Does it only apply to 1.14.3 clients? All villagers or just wandering traders / normal villagers? Do the villagers have a profession?

I still cannot reproduce this issue.

sachk:
I've had it happen on 1.13.2, 1.14.2 and 1.14.3. I'm using paperspigot, and it happened with and without the anticheat i'm using. I'm not using any other plugins that should affect villagers in any way. The villager trading GUI opens for around half a second, then closes. Relogging sometimes fixes it, as does a server restart.

Xuwznln:
Just look at the picture. I didn't type esc, it just close itself. and the villager is a librarian
http://i2.tiimg.com/505666/69870f0f2e6cbfb2.gif
and in the village, another villager is a farmer, the trade gui is normal.

PR #1135, opened 2019-07-22
Avoid desync acknowledge player digging

Brokkonaut:
Simply ignoring the AbstractAcknowledgePlayerDigging packet on older clients caused desyncs, we have to resend the block there.


Shevchik:
Instead of literally copying the block change single code, you need to create a static helper in blockcahngesingle impls that takes needed arguments and creates ClientboundPacketData and use it in both blockchangesingleimpls and ackplayerdig impls.

PR #1156, opened 2019-08-11
Fix block bug

EntryPointKR:
You can reproduce this bug in 1.13 or lower, If other player place a block(In my case, Blasting Furnace).
I checked it works fine at 1.13 and 1.5.2
I'm not sure that is this correct way for fix. Please review this.

[ChunkCache Usage](https://github.com/ProtocolSupport/ProtocolSupport/blob/d4126d64c1a8cf11e54b209ed1dc162ab7b5d5d8/src/protocolsupport/protocol/packet/middleimpl/clientbound/play/v_4_5_6_7_8_9r1_9r2_10_11_12r1_12r2_13/AbstractBlockChangeSingle.java#L26)


Shevchik:
Changing dimensions always resets the chunk data in client. So whatever your problem is, that is not the correct fix.

EntryPointKR:
Sorry for confusing, it can reproduce with SkinRestorer

Shevchik:
SkinsRestorer(from spigot) is known to break packet ordering and send garbage data, so we don't support any issues happening while running it.

EntryPointKR:
I see. Thanks for knowing me

PR #1199, opened 2020-01-02
NullPointerException

0ddlyoko:
In Minecraft 1.13.2, the method playerlist.getProfileBans().isBanned(mojangGameProfile) (Line 106) returns true if the player has been banned BUT doesn't check if it's expired.
The method playerlist.getProfileBans().get(mojangGameProfile) (Line 107) returns the ban BUT check if it's expired. If the ban is expired, it returns null.
Here, a NullPointerException is throw at line 132 ( https://prnt.sc/qin341 )
Minecraft has fixed this problem in 1.15 but that could be good to fix in ProtocolSupport 1.13.2


Shevchik:
We don't provide support for older versions. But checking for null there might still be worth it, just so it won't backfire in future versions.

Shevchik:
Implemented in https://github.com/ProtocolSupport/ProtocolSupport/commit/9db325e2d4f82e495bcab97d2f489f4deed269a8

Issue #1174, opened 2019-10-03
Brown mooshroom issue


djytw:
Red mooshrooms won't turn brown when struck by lightning.
Futhermore, if spawn a brown one without PS, then install PS and restart server, the brown one will turn red.

ps buildinfo: protocolsupport.ProtocolSupport$BuildInfo(buildtime: 2019.08.27 at 14:24:53 MSK, buildhost: build.true-games.org, buildnumber: 437, buildgit: https://github.com/ProtocolSupport/ProtocolSupport:origin/master:3a80c661fe665611a15d8cf39e91160e6775d2f2)

paper 1.14.4 build#208 (2019-10-03)

no other plugins

Shevchik:
Block mapping error i guess

djytw:
yes, only display as a red one. NBT tag is still type:brown

Shevchik:
Fixed by https://github.com/ProtocolSupport/ProtocolSupport/commit/0bf043fe915ff344ebc0fb4320552a039ba89c01

Issue #1137, opened 2019-07-22
Cats dont sit nor have tamed ocelot variants


MasterDash5:
Applies to 1.13.2 & lower

1.14.3: http://prnt.sc/oikwgl
1.13.2: http://prnt.sc/oikwmc

Shevchik:
Cat's are separate type in 1.14. I don't remap new type to old variants yet.

PR #1219, opened 2020-02-28
Some entities remapping

ChristophHD:
First of all, the majority of this remapping is due to things that I've seen by testing different clients (from 1.8.9) in a 1.15.2 paper server, using the latest development version which could be found at that moment (just some days ago). I've searched for that entities in the list of issues but not all of them were there, and I'm not sure if it'd have been better to create the issue and after that create the PR, anyway, I'm going to explain the things I've changed and why if needed.

- Rabbits were introduced in 1.8, but they were being remapped as chickens in that version.
- Added the brown mushroom remap ( from #1174 )
- Fixed the elder guardian remap (clients from 1.8 to 1.10 were seeing them like normal guardians, his metadata value used in previous versions to check if it was elder or not was functional just in older versions, before the entity was separated into a new kind of entity in 1.11, but using the same metadata field to check values about his spinnes, which previously also contained the elder guardian status, as I told earlier, but not since 1.11).
- Added legacy cat remap, I didn't write in the commit but it's from #1137, I've just assign them by the closest legacy color, although I've got some doubts about specific variants. It also shows the cats sit properly.
- Fixed a crash with arrows and tipped arrows for 1.13 clients (the shooter metadata value didn't exist until 1.13.1, so when a 1.13 client shoot an arrow or tipped arrow, it'd crash), as well as some tipped arrow colors fix (0 indicates no color in some versions, and it was changed to -1).

I guess that in some moment the default values used in some of the fixes could change as long as the minecraft protocol changes that values, making it non-functional in a future, but in 1.15.2, this seems to fix that problems. I've just tested all this changes from versions 1.8.9 to 1.15.2, including all his intermediate released versions.

### DIFF
--- a/src/protocolsupport/protocol/typeremapper/entity/EntityRemappersRegistry.java
+++ b/src/protocolsupport/protocol/typeremapper/entity/EntityRemappersRegistry.java
@@ -17,25 +17,7 @@ import protocolsupport.protocol.typeremapper.entity.metadata.types.base.AgeableE
 import protocolsupport.protocol.typeremapper.entity.metadata.types.base.BaseEntityMetadataRemapper;
 import protocolsupport.protocol.typeremapper.entity.metadata.types.base.InsentientEntityMetadataRemapper;
 import protocolsupport.protocol.typeremapper.entity.metadata.types.base.LivingEntityMetadataRemapper;
-import protocolsupport.protocol.typeremapper.entity.metadata.types.living.BatEntityMetadataRemapper;
-import protocolsupport.protocol.typeremapper.entity.metadata.types.living.BlazeEntityMetadataRemapper;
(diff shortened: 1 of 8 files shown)


Shevchik:
Since there wasn't any answer in a long time, i decided to cherry-pick the changes.

Currently partially implemented the changes in following commits:
1) Rabbits - https://github.com/ProtocolSupport/ProtocolSupport/commit/3eec19ef009d5ede03200af9ce49fec47fe454e5
2) MushroomCow - https://github.com/ProtocolSupport/ProtocolSupport/commit/0bf043fe915ff344ebc0fb4320552a039ba89c01
3) Arrows - https://github.com/ProtocolSupport/ProtocolSupport/commit/687bed7879fdd43c68e15287cc2d2776b34b35bf

Shevchik:
Implemented all the other fixes too.
Thanks for the help.

PR #1221, opened 2020-03-05
Update ProtocolSupport.java

UntouchedOdin0:
ye, lets not have that as a feature because I can't reload to get my plugins in.


Shevchik:
ProtocolSupport doens't support reloading. Removing command block will only do worse.

PR #1225, opened 2020-03-22
Legacy (Pre Flattening) items mapping update for colored beds in 1.12.x

WaveHD:
Here's a little fix for "`legacyitemtype.json`" and "`preflatteningitemid.json`".

With this configuration (obviously tested) players with `1.12`, `1.12.1` and `1.12.2` client versions, can see the colors of a bed item (as drop or as inventory element) even in a `1.15.2` native server.


Shevchik:
legacy item type is a generated file. All changes should be do to resource generator repo instead.

WaveHD:
Thank you for replying.
I apologize for not finding the correct repo first.

PR #1254, opened 2020-06-27
Cancel special-case block place packets from 1.8->1.9

lenis0012:
Basically, 1.9+ servers dont actually decode the cursor positions.
See: https://github.com/dmulloy2/PacketWrapper/blob/master/PacketWrapper/src/main/java/com/comphenix/packetwrapper/WrapperPlayClientBlockPlace.java

so, they cannot determine that a packet with -1 cursor values is a "special case" packet.
See: https://wiki.vg/index.php?title=Protocol&oldid=7368#Player_Block_Placement

This makes it to where 1.8 "special case" packets are processed as regular right clicks, which they are not.
This means that false fake player interact events get called, when the player is not actually right clicking.
For instance, when you finish eating, it sends 2 right clicks when you are holding food, where the first one is not a real interact packet.

### DIFF
--- a/src/protocolsupport/protocol/packet/middleimpl/serverbound/play/v_4_5_6_7/BlockPlace.java
+++ b/src/protocolsupport/protocol/packet/middleimpl/serverbound/play/v_4_5_6_7/BlockPlace.java
@@ -2,6 +2,7 @@ package protocolsupport.protocol.packet.middleimpl.serverbound.play.v_4_5_6_7;
 
 import io.netty.buffer.ByteBuf;
 import protocolsupport.protocol.ConnectionImpl;
+import protocolsupport.protocol.packet.middle.CancelMiddlePacketException;
 import protocolsupport.protocol.packet.middle.serverbound.play.MiddleBlockPlace;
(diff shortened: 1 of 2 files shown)


Shevchik:
According to the wiki you can also identify such packet by face being -1, so can we use that instead to do less operations?
Also same fix needs to be done for versions below 1.8 (v_4_5_6_7)

Shevchik:
Hmm, seems that middle packet already does special case -1 face, so it seems that either that special case needs to be removed or we need to find out in which case cursor positions are -1, but face is not.

lenis0012:
@Shevchik This is different.
When all data is set to -1, it means that the item you are holding should be updated.

That seems like a different thing than face -1 = interact
I am pretty sure that 1.9 clients do not send any packets when they are "done consuming" an item because that is now managed completely by the server.

lenis0012:
When you finish eating, the minecraft client tells the server to "update" the item by setting all bytes to -1.
this however, should not be interpreted as a right click which triggers another eat sequence, even though the client is not eating.

I imagine it also does weird things for buckets which update their state when emptied.

PR #1316, opened 2021-02-20
Fix legacy scoreboard prefix having a reset appended twice

alexsosnovsky:
This fixes an issue affecting users joining with 1.12.2 or earlier, to a server running the latest version with ProtocolSupport, where the last four characters of the scoreboard prefix were replaced with `§r§r`.

Before:

After:


Shevchik:
That's actually how it is supposed to work tho. The reset is inserted to reset the prefix color, since otherwise it would affect the suffix, or the entry itself.

alexsosnovsky:
I've modified it as such, but the original issue still persists which is that desirable characters of the prefix are getting cut off to accommodate the reset. The plugin managing the scoreboard actually handles the suffix color (whether it's an actual color or a reset), so the prefix color affecting the suffix isn't really a concern. Maybe the truncation could be avoided if the suffix already has its own color or reset at the start? If not, then what would be the best way to fix this issue? My goal is for the scoreboard to just look right.

Here's the scoreboard with the current commit:

Shevchik:
Leave the suffix empty and put everyting to prefix then. Make sure youy team entries arr color codes only, protocolsupport will then split the line to prefix and suffix for you.

alexsosnovsky:
Alright then I'll modify the plugin to work with how ProtocolSupport handles scoreboards.

In its current state, this pull request simply fixes the reset being appended twice, which is the extent of what I'm going to do with ProtocolSupport for now.

